PR #6, opened 2018-03-20
Should work™

marcjansen:
Please review.

### DIFF
--- a/src/components/TopLogo.vue
+++ b/src/components/TopLogo.vue
@@ -1,11 +1,11 @@
 <template>
 
   <v-avatar
-     :size={logoSize}
-     tile="false"
+     :size="logoSize"
+     :tile="true"
      class="wgu-top-logo"
     >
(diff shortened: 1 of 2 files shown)


src/components/TopLogo.vue
```diff
@@ -15,7 +15,7 @@
 export default {
   name: 'wgu-top-logo',
   props: {
-    logoSrc: {type: String, required: true},
+    logoSrc: {type: String, required: false},
```
chrismayer: IMHO a src should be provided, so required: false makes no sense, or?

src/components/TopLogo.vue
```diff
@@ -15,7 +15,7 @@
...
     logoSize: {type: Number, required: false}
```
chrismayer: A default value for logoSize would be nice. Maybe 100.

marcjansen:
@chrismayer, changed as requested.

Issue #29, opened 2018-07-23
Add geocoder widget


chrismayer:
A search widget to query a geocoder, like [Nominatim](https://wiki.openstreetmap.org/wiki/Nominatim), would a be a very useful component. 

The UI could be a textfield  with a result list below or a (possibly) existing combobox implementation , which provides the same functionality.

![image](https://user-images.githubusercontent.com/1185547/43084034-278067d0-8e98-11e8-9fc4-ac961caca9fe.png)

chrismayer:
Also [Photon](http://photon.komoot.de/) would be a great option to have as supported geocoder.

justb4:
I've started on this issue. Made Geocoders in OL before, but mainly struggling with the `Vue`, in particular the `Vuetify`-part for the component:

* template: I have a `v-btn` with `search` icon as toggle
* when enabled the searchinput box expands little bit [like this example](https://codepen.io/jreyesgs/pen/EWgBGM)
* toggle works ok, also aligns with Wegue `-btn` convention for toolbar
* next: what to choose as input box, thought [v-combobox](https://vuetifyjs.com/en/components/combobox) (as I used with ExtJS) but struggling receiving events (v-on:input, v-on:keydown etc trigger nothing) and popping up suggestions. Maybe should use other widget like regular `v-text-field`?

Below is a dump of the current component:

```
<template>

  <v-toolbar-items>
    <v-combobox
      v-model="selected"
      :items="searchResults"
      :label="label"
      append-icon=""
      :dark="dark"
      :persistent-hint="persistentHint"
      :hidden="showSearch"
      @change="onUpdate"
    ></v-combobox>
    
    <v-btn @click='toggle()' icon :dark="dark" slot="activator">
      <v-icon medium>{{icon}}</v-icon>
        {{text}}
      </v-btn>
  </v-toolbar-items>

</template>

<script>
  import {Mapable} from '../../mixins/Mapable';

  export default {
    name: 'wgu-geocoder-input',
    mixins: [Mapable],
    props: {
      color: {type: String, required: false, default: 'red darken-3'},
      icon: {type: String, required: false, default: 'search'},
      text: {type: String, required: false},
      label: {type: String, required: false, default: 'Nominatim Search'},
      dark: {type: Boolean, required: false, default: false},
      persistentHint: {type: Boolean, required: false, default: true}

    },
    data () {
      return {
        selected: '',
        showSearch: true,
        searchResults: [
        ]
      }
    },
    methods: {
[...]

PR #95, opened 2019-08-15
PR for #29 Geocoder widget

justb4:
A first rough version, bit in the dark on Unit testing, seems many existing tests fail because of config missing?

Implementation was still a bit more involved than foreseen, mainly because of some Vuetify `v-combobox` (dropdown state) quirks and dealing with the different Geocoder Providers behaviour.

Three Geocoders supported: `'osm'` (OSM Nominatim),  `'photon'` (Photon) and `'opencage'` (OpenCage, needs API key). We can extend with other later such as MapQuest and Bing. 

Multiple commits, but may be squashed.


src/components/geocoder/Geocoder.vue
```diff
@@ -0,0 +1,175 @@
...
+  <v-toolbar-items>
+    <v-combobox
+      return-object
+      :no-filter="noFilter"
+      v-model="selected"
+      :autofocus="autofocus"
+      :items="resultItems"
+      :label="placeHolder"
+      append-icon=""
+      :dark="dark"
```
chrismayer: By adding `:color="dark ? 'white': ''"` we could avoid the blue combobox, when having the focus. This would fit better to the general layout.

test/unit/specs/components/geocoder/Geocoder.spec.js
```diff
@@ -0,0 +1,35 @@
...
+describe('geocoder/Geocoder.vue', () => {
+  // Inspect the raw component options
+  it('is defined', () => {
+    expect(typeof Geocoder).to.not.equal('undefined');
+  });
+
+  describe('props', () => {
+    let comp;
+    beforeEach(() => {
+      comp = shallowMount(Geocoder);
```
chrismayer: Please look at `test/unit/specs/components/ol/Map.spec.js` or `test/unit/specs/components/measuretool/MeasureWin.spec.js` how to inject an app-config for the tests.

src/components/geocoder/GeocoderController.js
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * @constructor
+   * @param {String} providerName name of Provider.
+   * @param {Object} options config options for Provider
+   * @param {Object} parent callback parent
+   */
+  constructor (providerName, options, parent) {
+    this.options = options;
+    // Create Provider via Factory Method
+    this.provider = new PROVIDERS[providerName]();
```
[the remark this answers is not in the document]
justb4: Will issue `console.warn()` and return, ok? (Or is there a generic pop-up or so?)

src/components/geocoder/Geocoder.vue
justb4: Good! Could not find how to style/get rid of the blue CB!

src/components/geocoder/GeocoderController.js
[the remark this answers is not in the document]
chrismayer: `console.warn()` is totally fine, there is no generic popup.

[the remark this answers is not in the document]
justb4: done

test/unit/specs/components/geocoder/Geocoder.spec.js
justb4: Took some time (and sweat) but added more unit tests, in particular invoking the search input and combobox and verifying results even on map. Still could not get fake XHR and Timers working with `Sinon`, so faked with a `setTimeout()` in the test. So a real  "Nominatim" search is performed for a well-known place/address in Bonn...

Also merged with latest version upgrades from #104 like OL6, Vue, Vuetify, and still works.

Made a slide implementation change: replacing callback from the `GeocoderController`to the `Geocoder` with a real `async/await` approach. Made testing also easier.

So unless more tests are required this is done, possibly additional issues for enhancements.

src/components/geocoder/GeocoderController.js
```diff
@@ -0,0 +1,96 @@
...
+
+    // // Do the query via Ajax XHR, returning JSON. Async callback via Promise.
+    // json(ajax)
+    //   .then(response => {
+    //     // Call back parent with data formatted by Provider
+    //     this.parent.onQueryResult(this.provider.handleResponse(response));
+    //   })
+    //   .catch(err => {
+    //     this.parent.onQueryResult(undefined, err);
+    //   });
```
chrismayer: Can this be removed?

chrismayer:
I'll merge now. Thanks again @justb4!

PR #104, opened 2020-03-02
Update main deps

chrismayer:
Updates the main base libs to the current (non-breaking) versions:

- OpenLayers v6.2.1
- Vue v2.6.11
- Vuetify v1.5.23

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,14977 @@
+{
+  "name": "wegue",
+  "version": "1.0.0-dev",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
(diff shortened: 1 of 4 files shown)


justb4:
Do you plan to upgrade to Vuetify v2 (2.2.15 is now stable release)? Will have some impact (did the same in a much smaller Vue app, but new syntax is cleaner), but in longer term is better now that there are not too many widgets. Or could be done in separate PR. Even v3 (Vue3) is planned for this year.

chrismayer:
There are no plans so far about upgrading to Vuetify v2 and / or Vue v3 from my side. But if someone volunteers, why not...

JakobMiksch:
Tested it on my computer. All features seem to work like expected

chrismayer:
Thanks for the review, @JakobMiksch! I'll merge now.

justb4:
Also merged/pushed this my local branch ok.

PR #105, opened 2020-03-06
WFS layer support

chrismayer:
This adds support for WFS layers in the `LayerFactory`, so corresponding config objects in the `"mapLayers"` section of the app-config.json will be transformed to WFS layers, which are added to the map.

By setting a `"format"` property in the layer config an appropriate format will be loaded from the WFS (either `"GeoJSON"`, `"GML2"`, `"GML3`" or `"GML32"`).

By setting a `"loadOnlyVisible"` property in the layer config it is ensured the WFS data is only loaded for the visible BBOX of the map.


src/factory/Layer.js
```diff
@@ -26,7 +30,21 @@ export const LayerFactory = {
...
+    'GML32': GML32Format
+  },
+
+  /**
+   * Maps the format literal of the config to the corresponding WFS outputFormat.
+   * @type {Object}
+   */
+  wfsFormatMapping: {
+    'GeoJSON': 'application/json',
+    'GML2': 'text/xml; subtype=gml/3.2',
```
justb4: Think the format mapping for GML2 is incorrect?

src/factory/Layer.js
```diff
@@ -86,6 +106,61 @@ export const LayerFactory = {
...
+  /**
+   * Returns an OpenLayers WFS layer instance due to given config.
+   *
+   * @param  {Object} lConf  Layer config object
+   * @return {ol.layer.Tile} OL WFS layer instance
+   */
+  createWfsLayer: function (lConf) {
+    // set a default projection if not set in config
+    if (!lConf.projection) {
+      lConf.projection = 'EPSG:3857';
```
justb4: Shouldn't we assume the Map projection if no projection is specified for the WFS Layer?

src/factory/Layer.js
```diff
@@ -86,6 +106,61 @@ export const LayerFactory = {
...
+    }
+    // set a default WFS version if not set in config
+    if (!lConf.version) {
+      lConf.version = '1.1.0';
+    }
+    if (!lConf.format) {
+      lConf = 'GML3';
```
justb4: In my experience (GeoExt, Heron, but OpenLayers 2) if nothing is specified for WFS and GML version, the best chance of interworking is still WFS 1.1.0 with GML2. GML3 is/was often problematic. But things may have changed...

justb4: Plus: is meant at L125: `lConf.format = 'GML3'` else `lConf` is overwritten?

chrismayer: WFS 1.1.0 with its default outputFormat GML3 works quite well. See https://docs.geoserver.org/stable/en/user/services/wfs/basics.html

> Plus: is meant at L125: lConf.format = 'GML3' else lConf is overwritten?

:see_no_evil: oh yes, good catch.

src/factory/Layer.js
chrismayer: Yes you are right, using the map's SRS is more reliable.

chrismayer:
Thanks for your review @justb4! Very valuable comments. Very much appreciated! Would you be available to have another quick look over the changes? TIA!

JakobMiksch:
Just building WFS support for QGIS2Wegue: https://github.com/meggsimum/qgis2wegue/pull/12

I tried it with layers from [SDI Niederösterreich](https://sdi.noe.gv.at/at.gv.noe.geoserver/OGD/wfs) and it worked.

However this layer from [Salzburg Stadt](https://data.stadt-salzburg.at/geodaten/wfs?request=getcapabilities) gives an CORS error. But it works in QGIS.

```
    {
      "type": "WFS",
      "lid": "abfallsammelstelle",
      "name": "Abfall-Sammelstelle",
      "url": "https://data.stadt-salzburg.at/geodaten/wfs",
      "displayInLayerList": true,
      "visible": true,
      "format": "GML3",
      "typeName": "ogdsbg:abfallsammelstelle"
    },
```

Error message in Firefox:

image

Maybe there are some CORS settings we could adjust in OpenLayers

justb4:
Tested:

* when enabling the WFS layer at zoomlevel 0, `osm:osm-fuel` from https://ows.terrestris.de/geoserver/osm/wfs, about 100MB of Features are downloaded and my browser window freezes for some time. Ok, I should have zoomed in first, but maybe it is better to have a WFS with less features, or configure `maxFeatures` for that layer. This is always a problem with WFS...though there are strategies, like using `hits` output with clustering.
* the `projected` example also works, but also here loads of MB data (the Dutch admin borders are very long LineStrings, as I think they are aggregated from detailed Cadastral Parcel borders). So maybe here also a "smaller" WFS is better for UX. Though `maxFeatures: 10` in the config seems ignored?
* so it seems `maxFeatures` is ignored (?)
* suggestion for Dutch WFS with less features: National Parks (about 20 features), config tested:

```
    {
      "type": "WFS",
      "lid": "dutch-nat-parks",
      "name": "WFS (Dutch National Parks PDOK)",
      "url": "https://geodata.nationaalgeoregister.nl/nationaleparken/wfs",
      "typeName": "nationaleparken:nationaleparken",
      "version": "2.0.0",
      "maxFeatures": 10,
      "formatConfig": {
      },
      "format": "GeoJSON",
      "projection": "EPSG:28992",
      "loadOnlyVisible": true,
      "visible": false,
      "selectable": true,
      "style": {
        "strokeColor": "white",
        "strokeWidth": 2,
        "fillColor": "rgba(155,153,51,0.5)"
      },
      "attributions": "<a href='https://www.pdok.nl' target='_blank'>PDOK</a> by Dutch Kadaster"
    },
```
* tried if reprojection is supported, for example to enable a WFS that only supports WGS84 on a projected (`EPSG:28992`) Map. So setting `"projection": "EPSG:4326"` in the WFS Layer config. This should trigger reprojection in browser/OL (`Proj4.js`) (?). But somehow this does not work: the WFS request with "National Parks" is then:
https://geodata.nationaalgeoregister.nl/nationaleparken/wfs?service=WFS&version=2.0.0&request=GetFeature&typename=nationaleparken:nationaleparken&outputFormat=application/json&srsname=EPSG:4326&bbox=-209062.72000000003,362146.24,519062.72000000003,563853.76,EPSG:4326 
(`bbox` coords still `EPSG:28992`). Something similar was needed in [Geocoder](https://github.com/meggsimum/wegue/blob/master/src/components/geocoder/Geocoder.vue#L151).

chrismayer:
It seems like the WFS (Dutch National Parks PDOK) does ignore the feature limitation, see here: https://geodata.nationaalgeoregister.nl/nationaleparken/wfs?service=WFS&version=2.0.0&request=GetFeature&typename=nationaleparken:nationaleparken&outputFormat=application/json&srsname=EPSG:4326&count=10

justb4:
bbox in EPSG:4326 should work, this gives all 21 features:

https://geodata.nationaalgeoregister.nl/nationaleparken/wfs?service=WFS&version=2.0.0&request=GetFeature&typename=nationaleparken:nationaleparken&outputFormat=application/json&srsname=EPSG:4326&bbox=3,50,7,55,EPSG:4326

and this smaller bbox 3 features:
https://geodata.nationaalgeoregister.nl/nationaleparken/wfs?service=WFS&version=2.0.0&request=GetFeature&typename=nationaleparken:nationaleparken&outputFormat=application/json&srsname=EPSG:4326&bbox=3,50,5,52,EPSG:4326

(Ok, I see: OL WFS component emits the wrong request?...). Yes, also the new `PermalinkController` needs similar transformations when WGS84 center/extent is used on WebMerc.

> so it seems maxFeatures is ignored (?)

Hmm, strange behaviour:
https://geodata.nationaalgeoregister.nl/nationaleparken/wfs?service=WFS&version=2.0.0&request=GetFeature&typename=nationaleparken:nationaleparken&outputFormat=application/json&srsname=EPSG:4326&bbox=3,50,7,55,EPSG:4326&count=1
gives 21 Features.

and
https://geodata.nationaalgeoregister.nl/nationaleparken/wfs?service=WFS&version=2.0.0&request=GetFeature&typename=nationaleparken:nationaleparken&outputFormat=GML3&srsname=EPSG:4326&bbox=3,50,7,55,EPSG:4326&count=1

(`outputFormat` GML) gives 1 Feature! So only when requesting `application/json`.

We can address this at https://geoforum.nl (Dutch unfortunately). btw I worked on these national GDI services a long time (7y) ago...and developed a [love/hate relationship with WFS](https://justobjects.nl/2nd-time-around-wfs-v3-pygeoapi/) !.

chrismayer:
> What I think I meant is dat OL WFS client ignores maxFeatures?

Implemented in b93941f

> tried if reprojection is supported, for example to enable a WFS that only supports WGS84 on a projected (EPSG:28992) Map

Implemented in d5224b4 

So this ready, IMHO. Anyone available for another review? TIA!

JakobMiksch:
I tried it with these non-`EPSG:3857`  layers and works so far  without specifiy the CRS explicitly.
```
    {
      "type": "WFS",
      "name": "Hauptregionen NÖ 1:200000",
      "url": "https://sdi.noe.gv.at/at.gv.noe.geoserver/OGD/wfs?request",
      "lid": "hauptregionen_noe_1_200000",
      "style": {
        "strokeColor": "gray",
        "strokeWidth": 1,
        "fillColor": "rgba(20,20,20,0.1)"
      },
      "typeName": "OGD:HAUPTREGIONEN_200"
    },
    {
      "type": "WFS",
      "name": "WFS strange name_____üöäü?é",
      "url": "https://geodienste.hamburg.de/HH_WFS_Verkehr_opendata",
      "lid": "wfs_strange_name_ueoeaeue?e",
      "style": {
        "radius": 4,
        "strokeColor": "rgb(207, 16, 32)",
        "strokeWidth": 1,
        "fillColor": "rgba(207, 16, 32, 0.6)"
      },
      "typeName": "app:bikeandride"
    },
```

JakobMiksch:
referring to my recent comment. I get an error when I explicitly add the CRS of a non-`EPSG:3857` layer to the config:

```
{
    "type": "WFS",
    "name": "WFS Hamburg Bike",
    "url": "https://geodienste.hamburg.de/HH_WFS_Verkehr_opendata",
    "lid": "wfs_hamburg_bike",
    "style": {
      "radius": 4,
      "strokeColor": "rgb(207, 16, 32)",  
      "strokeWidth": 1,
      "fillColor": "rgba(207, 16, 32, 0.6)"
    },
    "typeName": "app:bikeandride",
    "projection": "EPSG:25832",
    "extent": [
      520461.116,
      5897482.201,
      614948.056,
      5985903.496
    ]
}
```
Error message:

image

I can fix that with adding this empty property:
```
"formatConfig": {
 },
```
Afterwards I get this error:

image


Did specify something wrong in my config file or is this an issue of Wegue?

justb4:
Think a projection definition needs to be added to the config, similar to the [EPSG:28992 Proj.4 def](https://github.com/meggsimum/wegue/blob/master/static/app-conf-projected.json#L25), something like

"projectionDefs": [
    [
    "EPSG: 25832",
    "+proj=utm +zone=32 +ellps=GRS80 +units=m +no_defs"
    ]
  ]

From: https://www.spatialreference.org/ref/epsg/25832/

chrismayer:
@JakobMiksch the missing `formatConfig` is now set automatically (introduced by 172e9da).

chrismayer:
> Think a projection definition needs to be added to the config, similar to the [EPSG:28992 Proj.4 def](https://github.com/meggsimum/wegue/blob/master/static/app-conf-projected.json#L25), something like
> 
> ```
> "projectionDefs": [
>     [
>     "EPSG: 25832",
>     "+proj=utm +zone=32 +ellps=GRS80 +units=m +no_defs"
>     ]
>   ]
> ```
> 
> From: https://www.spatialreference.org/ref/epsg/25832/

Note that it has to be "EPSG:25832" without a blank.

Changing that I see that the features are shifted to France, but should be placed in Hamburg. Is there something I am missing or has this something to do with #125?

chrismayer:
It seems that the format options for the OpenLayers GML format cannot be applied to the format itself (as it is doable for GeoJSON). It has to be applied to the `readFeatures` function call.

Compare https://openlayers.org/en/latest/apidoc/module-ol_format_GML3-GML3.html and https://openlayers.org/en/latest/apidoc/module-ol_format_GeoJSON-GeoJSON.html

See e30de1d

justb4:
I tested the branch. No problem with the 2 configs (default and `-projected.json`). But both WFS-es specify GeoJSON. This is not standard, and strictly GeoJSON only supports WGS-84. 
So I tried setting the `projected` config example to use `"format": "GML3",` for the Dutch WFS National Parks.  Got an error, see screenshot from debugger:

image

Strangely it (OL) can't find the Projection object for the `srsName` `http://www.opengis.net/gml/srs/epsg.xml#28992`. Though SRS `EPSG:28992` is registered.
This is typically WFS/GML mess: three ways (there is also a URN notation with `:`s) of specifying the same thing. One way out would be to register (with Proj.js) all variant-names for the same SRS. The different SRS naming-schemes have also long been used to imply axis order at least between OpenLayers and GeoServer.

Did not have time further. Maybe for now we can just support GeoJSON, though not all WFS-es support that, and strictly it is only WGS-84 (EPSG:4326).

JakobMiksch:
I tried again with my Hamburg layer:

```
{
  "title": "",
  "baseColor": "rgb(204, 0, 0)",
  "logo": "https://dummyimage.com/100x100/aaa/fff&text=Wegue",
  "logoSize": 100,
  "footerTextLeft": "",
  "footerTextRight": "",
  "showCopyrightYear": true,
  "mapZoom": 9,
  "mapCenter": [
    1125793.8189787832,
    7094088.201221157
  ],
  "mapLayers": [
    {
      "type": "WFS",
      "name": "WFS Hamburg",
      "url": "https://geodienste.hamburg.de/HH_WFS_Verkehr_opendata",
      "lid": "wfs_hamburg",
      "style": {
        "radius": 4,
        "strokeColor": "rgb(207, 16, 32)",
        "strokeWidth": 1,
        "fillColor": "rgba(207, 16, 32, 0.6)"
      },
      "typeName": "app:bikeandride",
      "projection": "EPSG:25832"
    },
```

I intentionally did not specify a global CRS, because I thought Wegue will transform the WFS layer then. But apparently it does not work. Here the error message:
image

chrismayer:
@JakobMiksch OpenLayers does not know about `EPSG:25832` by default. In case you want to use it it has to be specified in the Wegue config's [projectionDefs](https://github.com/meggsimum/wegue/wiki/Wegue-configuration#projectiondefs) section.

@

chrismayer:
> Maybe for now we can just support GeoJSON, though not all WFS-es support that, and strictly it is only WGS-84 (EPSG:4326).

I agree and we'll go for that. I could reproduce it and it has to be something of the "GML hell" you mentioned above. I'll merge as is and create an issue for GML3 with non-default SRSs and maybe add a comment to the readme as known restriction (or similar). Otherwise I have the feeling this and thus WFS-support will never be merged in. Let's start what we have (which is IMHO nearly 90%) and getting better step by step (instead if having nothing regarding WFS).

chrismayer:
Thank you all for your ongoing testing and your valuable feedback. Very much appreciated! I'll merge now.

Issue #111, opened 2020-03-12
Dynamic Layer configuration - Wegue format


justb4:
The current way to configure Layers is to define these in the Wegue config file like `static/app-conf.json`. 

But there are scenario's where a remote server provides an API Endpoint that supplies a "Layer List". OGC Standard examples are WMS Capabilities documents. The [GeoExt.tree.WMSCapabilitiesLoader](https://geoext.org/v1/lib/GeoExt/widgets/tree/WMSCapabilitiesLoader.html) is an example of a client that expands this to Layer objects.

And with the new OGC REST API an OpenAPI Feature "Collections" as a JSON doc, see [example here](https://demo.pygeoapi.io/master/collections?f=json). 

These should be addressed in separate issues. The issue here suggests a more direct approach and format: as Wegue Layers are already JSON nodes, a remote endpoint (or local file under e.g. `static`) could provide an **array of Wegue Layer definitions** like:

```
[
    {
      "type": "VECTOR",
      "lid": "dutch_windmills",
      "name": "Dutch Windmills",
      "url": "https://demo.pygeoapi.io/master/collections/dutch_windmills/items",
      "formatConfig": {
        "dataProjection": "EPSG:4326"
       },
      "format": "GeoJSON",
      "visible": true,
      "selectable": true,
      "style": {
        "iconUrl": "https://someiconserver.io/windmill2.png",
        "anchor": [0.5, 37],
        "anchorXUnits": "fraction",
        "anchorYUnits": "pixels"
      }
    },
    {
      "type": "VECTOR",
      "lid": "dutch_castles",
.
.
]
``` 

Such a Wegue Layer Def could be defined as a special Layer `type` with the endpoint URL and possibly some other parameters.  Something at minimum:

```
    {
      "type": "OSM",
      "lid": "osm-bg",
      "name": "OSM",
      "isBaseLayer": true,
      "visible": true,
      "displayInLayerList": false
    },
    {
      "type": "WEGUELAYERCOLLECTION",
      "url": "https://some.server.io/api/layer_list"
    },
```

chrismayer:
So, thanks for this @justb4. One general question: Is this all about "divide and conquer", so that we separate the `mapLayers` from the application config? Or is the main motivation the preparation of embedding remote endpoints as you described for OGC REST API an OpenAPI Feature? Are there any other advantages? 
[...]

PR #113, opened 2020-03-13
#111 dynamic layers in Wegue format

justb4:
Introduces a new Wegue Layer type `LAYERCOLLECTION` which is mainly a URL pointing to a remote endpoint (or file local to the Wegue App) that contains an array of Wegue Layers (which in theory could even again contain `LAYERCOLLECTION`s). 

* As the loading of the URL (via `fetch()`) is asynchronous, several functions are now `async/await` (ES2015 syntax). 
* test added, NB final version needs to point to Wegue GH link

Later similar "expanding layer types" could be added for WMS, WFS etc (via GetCapabilities) and even OGC OpenAPI Features ("WFSv3").


justb4:
@chrismayer sorry, your questions escaped me.

ad 1) `test/unit/specs/ol/Map.spec.js` was a duplicate of `test/unit/specs/components/ol/Map.spec.js`.

ad 2) `me = this` is uneccessary in ES2015 with arrow functions

Ok strictly not part of issue, but cleanup which I considered too minor for issue/PR.

chrismayer:
Thanks for clarification.

One further point: Due to asynchronous handling the position in the `mapLayers` array is not resprected for type `LAYERCOLLECTION`. In the app-conf.json the demo layer for a `LAYERCOLLECTION` is laced somewhere in the middle of the `mapLayers` array but is rendered topmost in the layer stack.

justb4:
Hmm, yes. The Layers coming from the `LAYERCOLLECTION` are reversed as to have them internally added bottom to top just like the regular Layers. But yes, the async nature could lead these always added as last/topmost. Tough problem to solve, if ever a problem in practice: 1) usually the entire app is configured from a single `LAYERCOLLECTION` or the latter is a series of overlay Layers on top of regularly configured baselayers (Geolicious use case).

Solution would be to "stack" any remaining regular Layers (not adding them) after the `LAYERCOLLECTION` and add them after the `async` finishes. Would be interesting with multiple `LAYERCOLLECTION`s....

As an alternative, for progress, we could merge this PR and open a new issue for "`LAYERCOLLECTION`  messes up Layer stack ordering"...

PR #119, opened 2020-03-19
XYZ layer: add attribution config

JakobMiksch:
Adds the attributions parameter to the XYZ layer configuration. Also adds a example to the default config file.


static/app-conf.json
```diff
@@ -85,6 +85,16 @@
...
     },
 
+    {
+      "type": "XYZ",
+      "name": "OpenTopoMap",
+      "url": "https://tile.opentopomap.org/{z}/{x}/{y}.png",
+      "attributions": "Map data: <a href=\"https://openstreetmap.org/copyright\">©OpenStreetMap</a>-contributors, SRTM | Map representation (Kartendarstellung): © <a href=\"http://opentopomap.org/\">OpenTopoMap</a> (<a href=\"https://creativecommons.org/licenses/by-sa/3.0/\">CC-BY-SA</a>)",
+      "lid": "opentopomap",
+      "displayInLayerList": true, 
+      "visible": true
```
chrismayer: I'd prefer to have the OpenTopoMap layer not visible per default since it overlays the OSM base map. What do you think?

JakobMiksch: yes, sure.

Issue #38, opened 2018-12-11
Add map permalink functionality


chrismayer:
Goal is to add a simple permalink functionality for the map state like it is done in http://openlayers.org/en/latest/examples/permalink.html

justb4:
We are looking into this in a project. I paste some comments here, as there are implementation options.

First there is map basic context as in
https://openlayers.org/en/latest/examples/permalink.html#map=12.659486938142972/1107025.7/7011582.75/0

But the encoding is somewhat specific, would require specific parsing (of '/' separated) and one always needs a complete spec including `rotation`.

Two real-world examples:

*  France national geoportal:
https://www.geoportail.gouv.fr/carte?c=4.527906651375692,44.98704671209671&z=14&l0=GEOGRAPHICALGRIDSYSTEMS.MAPS.SCAN-EXPRESS.STANDARD:WMTS(1)&permalink=yes

* Swiss topo portal:
https://map.geo.admin.ch/?topic=swisstopo&lang=en&bgLayer=ch.swisstopo.pixelkarte-farbe&E=2560256.11&N=1256280.28&zoom=4&layers=ch.swisstopo.zeitreihen&catalogNodes=1392&layers_timestamp=18641231&swipe_ratio=0.56

These also have sharing (buttons) options (link, embed iframe) possibly separate issue. 
Important is that both examples above are  not using the location hash (`#`) but query (`? &`) params. 

I've implemented location hash before, example:
https://app.map5.nl/nltopo/#osm/opentopo/13/5.758038/52.31908

One big disadvantage (over query params) I found (also in the OL example above)  is that the browser back-button and history is "screwed up", i.e. goes through previous map-contexts i.s.o. previous web-page. (Or I missed something there).

Maybe both options are possible as a config option when at least the permalink format is standardized, I propose: `<?|#>c=<x>,<y>&z=<zoom>&r=<rotation>` as minimum. More extended implementations may add Layers, projection, popups,tabs etc. In all cases we can use `URLSearchParams` to parse either `location.hash` or `location.search` based on config option. I've even seen implementations that have the parameter names as config option, or a param-prefix like `map_` to get `map_z` etc, as to avoid collisions with server-side or other components (like vue-router). The prefix with default empty string is a good option.

dkerkow:
This sounds reasonable to me. Just one thought that I would like to add:

What about using the map bounds instead of the map center?

[...]

PR #121, opened 2020-03-23
PR for #38 permalink

justb4:
This is a minimal permalink implementation, it is configuration-driven. Using a new `permalink` config object, example:

```
  "permalink": {
    "location": "hash",                  # how to specify in address bar hash(#) default or query(?)
    "projection": "EPSG:4326",   # projection for permalink coords (default: map projection)
    "paramPrefix": "",                   # prefix for permalink parameters (default: empty str)
    "history": true,                        # should history be kept and shown in browser history (default: false)
    "precision": 6                          # for numbers: center, zoom, extent
  },
```
Defaults apply, so `"permalink": {}` is also valid.

- [x] Encoding: using `c` (center), `z` (zoom) , `r` (rotation) with optional prefixes e.g. `map_c=`. 
- [x] Example: http://localhost:8081/#z=1.9867&c=2.306,39.5258&r=0
- [x] Non-map related params, e.g. `appCtx=` are preserved
- [x] optional history `popstate` capture
- [x] optional parameter prefixes: `map_z=1.9867&map_c=2.306,39.5258&map_r=0`
- [x] optional projection in permalink, e.g. use WGS84 i.s.o. OSM Merc coords
- [x] Implementation in new class `src/components/ol/PermalinkController.js` as ordinary class
- [x] Allow for overriding in `Map.vue/PermalinkController.js` app-specific subclasses via Factory Method.
- [x] prepared for URL sharing (`PermalinkController.getParamStr ()`) method
- [x] prepared for IFrame Embed sharing (`PermalinkController.getEmbedHTML ()`) method
- [x] allow map `extent` URL parameter i.s.o. `center+zoom` (configurable)
- [x] allow layers (`lid`s?) to be encoded (challenge: layers are loaded `async`) 
- [x] Unit Tests


justb4:
All planned permalink features now implemented. Need to revisit layer handling when #113 is merged (async layer creation).

justb4:
So this was already there, only unit tests are lacking. The `ShareButton` component (#128) builds on this PR.

justb4:
Should be good to go.

src/factory/Layer.js
```diff
@@ -59,8 +59,9 @@ export const LayerFactory = {
   getInstance (lConf, olMap) {
     // apply LID (Layer ID) if not existent
     if (!lConf.lid) {
-      var now = new Date();
-      lConf.lid = now.getTime();
+      // Make a unique layerId from Layer name and URL so contexts
+      // like permalinks can be reapplied.
+      lConf.lid = btoa(lConf.url + lConf.name).substr(0, 6);
```
chrismayer, line 64: Does every layer has a property `url`? Not sure about that

chrismayer:
Looks good to me, thanks @justb4. One little question above and one general question:

Will it be easily possible to aplly another permalink structure? I have a public application called [Shareloc](https://apps.meggsimum.de/shareloc/), which provides shareable links for OpenLayers maps. In the future I might want to upgrade this to Wegue. The permalink in this PR looks a bit different from the one used with Shereloc (e.g. `https://apps.meggsimum.de/shareloc/share.html?zoom=7&X=8.541570112005266&Y=50.5918788090799&bgLayer=osm.base`). So I would need to map / modify the parameter keys and maybe add more. Would this easily possoble with the approach in this PR? TIA!

justb4:
Not directly, though there is some abstraction already via the config: `paramPrefix` which is reflected in the permalink param names and their encoding/decoding. This was somewhat similar as in GeoExt. 

A next step would be to make all param-names configurable with sensible default. Not too difficult as parameter naming is just in a few places (2). Only `bgLayer` is a bit specific in the Shareloc approach. In the current approach there is just `layers` as some Layers are already background-type.

chrismayer:
> Not directly, though there is some abstraction already via the config: `paramPrefix` which is reflected in the permalink param names and their encoding/decoding. This was somewhat similar as in GeoExt.
> 
> A next step would be to make all param-names configurable with sensible default. Not too difficult as parameter naming is just in a few places (2). Only `bgLayer` is a bit specific in the Shareloc approach. In the current approach there is just `layers` as some Layers are already background-type (`isBaseLayer`).

Thanks for clarification.

chrismayer:
I just tested the Permlink behaviour and found something unexpected. If a layer gets activated it is not added to the permalink. Steps to reporduce:

- Open http://localhost:8081/?appCtx=projected
- Activate ` WMS - Natura 2000 Areas - PDOK` layer with the LayerList tool.
- Check the URL it still remains as before.:

grafik

justb4:
Confirmed. As soon as you alter the Map View (pan, zoom) the Layer is added. Almost sure the problem is that changing active Layers does not trigger `moveend` from the OL Map (to which `PermalinkController` listens). Investigating. May need to listen to other generic Event, `change`?

chrismayer:
It is not intended that activating a layer triggers `moveend` on the map. I think listening to `change:visible` on the map's [LayerGroup](https://openlayers.org/en/latest/apidoc/module-ol_layer_Group-LayerGroup.html) should be the correct event. Derivable by [map.getLayerGroup()](https://openlayers.org/en/latest/apidoc/module-ol_Map-Map.html#getLayerGroup).

justb4:
Yes, aware by now. Approach now (before seeing comment above) is to listen to two additional events: `'change:visible'` for each Layer and `'change:length'` for the Map's `Layer Collection`. The latter is needed whenever Wegue support dynamic Layer addition. Also added tests and this approach works. Tried listening to `change:visible` for Map's `LayerGroup` but this had no effect. But I was not familiar with LayerGroup.

I will push the changes for now and make it WIP for review.

chrismayer:
Thanks for your ongoing work @justb4 ! LGTM, I'll merge now.

Issue #126, opened 2020-04-24
Provide sizing in Image/Icon Vector Style Layer config 


justb4:
Wegue Vector Layers can be configured with `Style` definitions. These are mapped to OpenLayers `Style` objects by the Wegue [OLStyle.js](https://github.com/meggsimum/wegue/blob/master/src/factory/OlStyle.js) Factory.

For Icon `Styles` one can configure attributes like `iconUrl`, `anchor` etc. But there is currently no means to influence sizing (e.g. for SVG Icons), so icons get their original size on the Map. 
More advanced would even be to size icons with the Map's Zoom level, but let's first keep it simple: allow a `size` or `scale` parameter in the Style config, for example:

```
    {
      "type": "VECTOR",
      "lid": "dutch_castles",
      "name": "Dutch Castles",
      "url": "https://demo.pygeoapi.io/master/collections/dutch_castles/items",
      "formatConfig": {
        "dataProjection": "EPSG:4326"
       },
      "format": "GeoJSON",
      "visible": true,
      "selectable": true,
      "style": {
        "iconUrl": "./static/icon/castle2.png",
        "scale: 2",      <===========================
        "anchor": [0.5, 37],
        "anchorXUnits": "fraction",
        "anchorYUnits": "pixels"
      },
      "styleSelected": {
        "iconUrl": "./static/icon/castle2-select.png",
        "scale: 2",    <===========================
        "anchor": [0.5, 37],
        "anchorXUnits": "fraction",
        "anchorYUnits": "pixels"
      }
    },
```

NB `size` did not seem to have any influence when tried in  [OLStyle.js](https://github.com/meggsimum/wegue/blob/master/src/factory/OlStyle.js), but `scale` has.
So `scale` seems best option now. A good first issue!

There are probably more parameters, this is also a bit why we are in favor of a [JavaScript Wegue config](https://github.com/meggsimum/wegue/issues/123).

chrismayer:
`scale` seems to applied already. So can we close this?

![image](https://user-images.githubusercontent.com/1185547/101767160-2012f580-3ae4-11eb-966a-593732853aef.png)

justb4:
Yes, closing...

PR #127, opened 2020-04-24
#126 allow scale parameter in Vector Layer Icon Style config

justb4:
Simple fix, to allow scale in Style Icon config.


chrismayer:
Thanks for this. I'll merge now.

Issue #98, opened 2020-01-30
Config-file location


JakobMiksch:
Currently the config file must be placed in the `static` folder, whereas the rest of application-specific code is located in `app`. 

Therefore, it would be clearer to also move the config file to the `app` folder. Ideally the `init:app` script should create this initial config file. So, the user can immediately start the application.

justb4:
Agree, there is also issues with merging upstream into a clone. Possibly `main.js` could first look in `app` and then `static` for a config as a saw in @chrismayer branch: https://github.com/chrismayer/wegue/blob/app-fldr-conf/src/main.js#L47.

Placing a starter config under `app/` could be also something `init:app` could do...

chrismayer:
Yes, the development in https://github.com/chrismayer/wegue/blob/app-fldr-conf/src/main.js#L47 is based on this issue. I should have pulled it to me. Done by now :wink: 

 > Placing a starter config under app/ could be also something init:app could do...

That was the plan as described by @JakobMiksch in the initial description of the issue: "Ideally the init:app script should create this initial config file. So, the user can immediately start the application"

justb4:
Think we can close this one.

PR #131, opened 2020-05-24
PR for #98 app-specific config file location under app/

justb4:
Solves #98 :

* on startup `main.js` will first search in `app/static/` and then `static/` for th app context config .json file
* works for both dev-server and built (prod) versions

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -33,17 +33,27 @@ if (appCtx) {
   appCtxFile = '-' + appCtx.replace(/(\.\.[/])+/g, '');
 }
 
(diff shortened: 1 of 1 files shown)


chrismayer:
Thanks @justb4, LGTM. Makes things more flexible. I'll merge now.

As already discussed offline it looks pretty the same as https://github.com/chrismayer/wegue/blob/app-fldr-conf/src/main.js :wink:

justb4:
@chrismayer Thanks. Yes, similar, but wait until you see PR #132...

Issue #130, opened 2020-05-24
Application config and directory restructure


justb4:
This issue relates/depends on to #98 but goes a bit further. 

A typical Wegue project will clone the Wegue GH repo, add both new custom (Vue) components and/or modify core Wegue code and components. Currently this implies changes to files under the `src/` dir. Merging-in changes from upstream Wegue GH repo later gets more and more tedious and error-prone.

Ideally we would not need to touch any files under `src/` but develop app-specific components and other resources under the `app/` tree. This should also be a step towards making Wegue a real app-framework with a CLI to create a starter-app under any other dir/ repo i.s.o. within a clone of the Wegue repo under `app/`. The advantage of this process though is that one constantly have all files at disposal, also e.g. within debuggers.

Via this issue I propose the following changes:

* remove `app/.gitignore`
* place `app-conf.json` under `app/static` dir (via #98)
* implement/merge [the simple solution](https://github.com/chrismayer/wegue/blob/app-fldr-conf/src/main.js#L47) which first first looks for app config under app-dir.  
* put app-specific components and if possible modified Wegue components under `app`-dir
* make an `app/components/` shadow dir with app-specific/Wegue-app-specific-modified components
* make `app/static` a shadow dir from `static/`: with app-specific resources (e.g. css, icons etc) that overwrites `static` in builds
* reference both `app` and `src` Vue components (requires build config changes)
* extend Wegue CLI `npm run init:app`

For the latter the CLI `npm run init:app` needs to copy the following Wegue  files under `app`-dir: `WguAppTemplate, AppFooter, AppHeader`. These will be app-specific anyway at some point.

Note that we already successfully applied this restructuring within the Geolicious project for NPLH. Need to see what is the best strategy to pull-in this work via https://github.com/Geolicious/wegue. 
That huge change costed little dev-time, and helps with future Wegue dev-workflows, much the way e.g. many template-systems like Django work (shadowing a subset of files with a searchpath).

justb4:
Think we can close?

chrismayer:
Yes, I'll close it

PR #132, opened 2020-05-24
For #130 app restructuring

justb4:
This provides a new way to develop a Wegue app as a step towards a full framework/CLI version ala Vue and React. The design here has been proven to work in the Geolicious NPLH project. The main addition here is to add a basic CLI to get a "starter app". The current workflow will be

* clone Wegue repo (or version)
* `npm install` to install dependencies
* `npm run init:app` : this will now create an app under `app` as a copy of `app-starter` dir
* `npm run dev` will still run the default app etc.
* ditto for the prod build (`npm run build`)

NB a second `npm run init:app` will not overwrite existing files!
Now under `app/` the backbone of a Wegue app is present:

* main app components (.vue files): `WguApp`, `WguAppTemplate` and the `WguAppHeader/Footer` 
* config files under `app/static`
* other resources like icons, css, data and fonts can be placed under `static`

A developer can now adapt all files under `app`. 

TODO: we may consider moving `src/main.js` also under starter-app.


justb4:
@chrismayer Should be good to go.

app-starter/README.md
```diff
@@ -0,0 +1,8 @@
+# Wegue App Dir
```
chrismayer, line 1: Shouldn't this be placed in `app` instead of `app-starter`? Or is this here because this file is copied by `init:app` script?

justb4: Yes, good suggestion. The intention was that it would be the README for the app dir. That may be somewhat confusing placing it in `app-starter`. But as nothing is overwritten during copy, I can move it to `app` and have a separate README in `app-starter`. Hold on...

justb4:
Ok added and modified READMEs.

chrismayer:
Very cool. Thanks for your ongoing work @justb4. I'll merge this now.

Issue #106, opened 2020-03-06
Add WFS layer support


chrismayer:
Add support for WFS layers in the `mapLayers` section of the application configuration.

JakobMiksch:
documentation of the WFS config is also required

chrismayer:
Implementation done by #105.

Documentation still missing, moved this to a separate issue #148.

PR #136, opened 2020-07-01
Docsify Website for Wegue

JakobMiksch:
Preview (maybe not up to date): https://jakobmiksch.github.io/wegue-docs/#/

- This is a basic setup for a  website for Wegue based on [Docsify](https://docsify.js.org/#/). It consists of a landing page and the documentation of Wegue. 
- The website does not need to be built, the HTML will be generated on the fly. Hence it can easily be shared with GitHub.
- The current documentation of this PR is from the [Wegue Wiki](https://github.com/meggsimum/wegue/wiki)

Website:
- [x] GIF on landing page has 3MB and might be too big  --> compressed to 1.5 MB
- [x] screenshots of example websites are relatively large --> compressed significantely
- [x] add missing documentation (e.g. WFS) --> move to #106 

Launch: 
- [ ] change basePath
- [ ] activate GitHub pages --> publish `docs` directory and link it to [www.wegue.org](http://www.wegue.org)


chrismayer:
This is cool. Thanks a lot @JakobMiksch! Should ths be merged and then the open points will be addressed in (a) separate PR? Or would you like to take care of the the remianing points first?

JakobMiksch:
Great. 
I think I will make some more adaptations and then we could merge this.

JakobMiksch:
currently the demo webmap is published with GitHub Pages. Publishing the docs as a separate website in this repo will probably not work.
I tried moving the `docs` into the `demo` directory and vice versa. But both does not work. I guess because the dynamic Javascript magic in from `docsify` doesn't like that ... 

@chrismayer do you have any preference how to deal with that?

justb4:
I think it should be possible to publish to `docs`. Though given the dynamic HTML generation I wonder where the magic happens. Strictly `gh-pages` is no longer required, one can use `docs`, but then some settings are required for the repo which only @chrismayer can do. One of them is also DNS/CNAME to map site to wegue.org domain. See 
https://docsify.js.org/#/deploy . 
If really a command (`docsify-cli`?) is required, GitHub Workflows/Actions work well these days. We could use that for Dockerfile publication as well, see e.g. [this example](https://github.com/justb4/docker-mapproxy/blob/master/.github/workflows/main.yml). Or good old Travis.

If there is really a problem we could consider a static site/HTML generator which works from `.md` files. In the Vue.js world [VuePress](https://vuepress.vuejs.org/) is popular.

chrismayer:
I think the general approach to have the files in the docs folder is good, since github allows to publish directly from master/docs folder. I just have to activate this. What I also have to do afterwards is to re-configure Travis, so it does not push the demo-application to the gh-pages branch. Maybe it can be pushed to master/docs/demo or something similar.

JakobMiksch:
Added more content 96572026583b12faadde067147149bc215f5afb8
There is no preview online. But you pull the changes and start a local web browser in the `docs` directory

chrismayer:
Thanks @JakobMiksch ! Looks really good. I think this is good to. Once merged I'll activate that the docs folder is used as source for the GitHub Pages site.

chrismayer (approved):

PR #139, opened 2020-09-02
website updates config section

justb4:
* fix link to  `mapprojection` section
* various rephrasing
* added proj4-related links
* tilegrid config definition
* remove Projection and Tilegrid config from example config (they are not used in the config furhter)
* use 2 examples: the one from the demo and link to one with projections


chrismayer:
LGTM, thanks @justb4 ! I'll merge now.

PR #143, opened 2020-09-10
created geolocation button and functions

juergenhah:
Component that uses HTML5 Geolocation API to show device location on the map
Features:
- use Html5 geolocation functionality to receive geolocation
- show a marker on the map of actual geolocation
- zoom the map to the actual geolocation
- adds a new layer to the layer tree to show/hide the geolocation marker
- adds a button in the toolbar with the locating symbol

added the functionality to the 'app-starter' kit


chrismayer:
Thanks for your contribution @juergenhah. I just tested and it works quite well. 

Some thoughts: For me it looks like the position of the geolocation API is not updated automatically. I have to push the button every time I want to see my current location. This could be a use-case but IMHO most of the users would expect to activate a toggle-button and get the position updates automatically as long as the toggle button is pressed.  
Are you aware of [ol/Geolocation](https://openlayers.org/en/latest/apidoc/module-ol_Geolocation-Geolocation.html)? This allows to watch the position of the geolocation API. In case the you don't want to track position there is a function [setTracking](https://openlayers.org/en/latest/apidoc/module-ol_Geolocation-Geolocation.html#setTracking), which can be set to false.

What do you think @juergenhah ?

juergenhah:
Hi @chrismayer ,

you are right, the position is only set once. My use case was to zoom to the users location, such that the user can see items of other geodata layers that are next to the position. However, I agree to your suggestion to implement an automatic update. I will commit the new solution sooooon... 

Thanks in advance and looking forward to your reply.

chrismayer:
Superseded by #150 due to mixed up history.

PR #147, opened 2020-10-14
Add .dockerignore file

chrismayer:
Adds `.dockerignore` file so the build context reduces massively.

Please review @justb4 ... TIA!


justb4:
Yes, good idea, especially ignoring `node_modules`, speeds up build. Good to go!

chrismayer:
Thanks for the review @justb4.

Issue #149, opened 2020-10-16
Add tracking to geolocator component


chrismayer:
The newly introduced Geolocator component (see #143) allows to show the user's position on the map by a button click.
Most of the users would expect to activate a toggle-button and get the position updates automatically as long as the toggle button is pressed (tracking).
Therefore OpenLayers provides [ol/Geolocation](https://openlayers.org/en/latest/apidoc/module-ol_Geolocation-Geolocation.html) as a wrapper for the HTML Geolocation API, which allows to watch the position of the Geolocation API easily. 
So in order to support "tracking" the Wegue Geolocator component should be refactored, so it uses ol/Geolocation. 
In case the you don't want to track position there is a function [setTracking](https://openlayers.org/en/latest/apidoc/module-ol_Geolocation-Geolocation.html#setTracking), which can be set to false.

PR #150, opened 2020-10-16
Add Geolocation button component

chrismayer:
This PR adds a Geolocation button component as provided in #143.

Due to some conflicts in the history the original PR is screwed up and this provides the original work of @juergenhah with some little optimizations by me in a cleaned history. So most of the work done by @juergenhah (and all kudos).

Adding the original description of #143 here for completeness:

Component that uses HTML5 Geolocation API to show device location on the map Features:

  - use Html5 geolocation functionality to receive geolocation
  - show a marker on the map of actual geolocation
  - zoom the map to the actual geolocation
  - adds a new layer to the layer tree to show/hide the geolocation marker
  - adds a button in the toolbar with the locating symbol
  - added the functionality to the 'app-starter' kit

Optimizations:

  - make zoom animation configurable
  - Get rid of `setTimeout`
  - consequent use of ES6 notations (const/let, arrow funcs ...)
  - Some naming / typos

Some follow-up work should be done as stated in #149.

### DIFF
--- a/app-starter/components/AppHeader.vue
+++ b/app-starter/components/AppHeader.vue
@@ -64,6 +64,7 @@ import MeasureToolToggleButton from '../../src/components/measuretool/ToggleButt
 import InfoClickButton from '../../src/components/infoclick/ToggleButton'
 import ZoomToMaxExtentButton from '../../src/components/maxextentbutton/ZoomToMaxExtentButton'
 import Geocoder from '../../src/components/geocoder/Geocoder'
+import Geolocator from '../../src/components/geolocator/Geolocator'
 
 export default {
   name: 'wgu-app-header',
@@ -73,7 +74,8 @@ export default {
     'wgu-layerlist-btn': LayerListToggleButton,
     'wgu-helpwin-btn': HelpWinToggleButton,
     'wgu-measuretool-btn': MeasureToolToggleButton,
-    'wgu-infoclick-btn': InfoClickButton
+    'wgu-infoclick-btn': InfoClickButton,
+    'wgu-geolocator-btn': Geolocator
   },
   props: {
     color: {type: String, required: false, default: 'red darken-3'}
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -185,6 +185,10 @@
     "wgu-helpwin": {
       "target": "toolbar",
       "darkLayout": true
+    },
+    "wgu-geolocator": {
+      "target": "toolbar",
+      "darkLayout": true
     }
   }
 }
--- a/src/components/geolocator/Geolocator.vue
+++ b/src/components/geolocator/Geolocator.vue
@@ -1,0 +1,108 @@
+<template>
+<span>
+   <v-btn @click="geolocateUserAndShowMarkerOnMap" icon dark>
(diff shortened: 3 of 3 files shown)


justb4:
Very useful addition! Two main comments:

* Could the Geolocator Button be made an on/off toggle? See e.g. this example: https://app.map5.nl/nltopo
* Support Projections (now EPSG:4326 is assumed). e.g. running the app with `appCtx=projected` does not show proper location. The solution is similar as in `Geocoder`, using `proj.js` see e.g. [these lines in Geocoder.vue](https://github.com/meggsimum/wegue/blob/master/src/components/geocoder/Geocoder.vue#L149).

Minor comments:

* Several used-once functions like `createAndRemoveExistingLayer()` and the `Style` are defined outside the Vue template. They could be under `methods`.
* Style is fixed to `blue` and `person_pin_circle`. Can this be made configurable in `props`, with those values as defaults?
* "Keep Map Centered on Location" is often dependent on type of application: when the user moves should the map remain centered? Often, when a user pans the map centering is stopped. These are typically configurable options (`props`) with defaults.

chrismayer:
Thanks for your review @justb4. I addressed the following things:

- Support projections
- Props for marker icon color and text are now configurable by props
- Moved the external functions into the Vue component

For your point `Could the Geolocator Button be made an on/off toggle?` a separate issue (#149) exists, so this will be tackled in a follow-up.

Is this good to go?

justb4:
Some major comments:

* the component uses `navigator.geolocation.getCurrentPosition()`. Don't know if this is the intention but this fetches location only once. Usually one would want to use tracking, in that case `navigator.geolocation.watchPosition()` (and `clearWatch()`) should be used. See [1] and an OpenLayers example [2]
* for each location the entire geolocation Vector Layer is removed and added. Only the source Feature needs to be refreshed, or could even only be moved.

Suggestion: OL provides a GeoLocation helper class [3]. Isn't that class handier to use? Also for event-based for later other components in a Wegue app.

Other:
* it is usually an option to keep position in the middle, follow/tracking mode. Only if the user pans the map is this disabled. We can make these two a config option: `autoCenter` (center when location found) and `lockCenter` (always keep position centered). There is usually a whole range of config options, with sensible defaults for a GeoLocation widget. See for example this Leaflet Location control. [4]

[1] https://www.w3.org/TR/geolocation-API/
[2] https://openlayers.org/workshop/en/mobile/geolocation.html
[3] https://openlayers.org/en/latest/apidoc/module-ol_Geolocation-Geolocation.html
[4] https://github.com/domoritz/leaflet-locatecontrol

Minor:

* the `created()` function can be simplified. (outer `this` is always available in JS arrow functions, the `let me = this` pattern is never needed.)

```
  created () {
    WguEventBus.$on('ol-map-mounted', olMap => {
      this.map = olMap;
    });
    this.isGeolocationAPIAvailable = navigator.geolocation;
  },
```

chrismayer:
Thanks again for your valuable feedback @justb4. Most of your points are already listed in  #149, which should be addressed quite directly after merging this as follow-up. Therefore I'd pledge to merge this as is.

justb4:
You were just ahead, but approach fine with me!

Issue #157, opened 2020-12-16
Consider enhanced templating for map hover and vector labels


chrismayer:
Currently the content of the hover popup and the labels for vector layers on the map are based on single attributes of the underlying data. A more sophisticated templating mechanism could be useful here. So a string like 

`Name: {{name}} / Foo: {{foo}}`

could be the base for the hovering text and the vector labels.

JakobMiksch:
fixed by #269

PR #158, opened 2020-12-16
Add basic support for vector labels

chrismayer:
This adds basic support to configure labels in the style for vector based layers. Base for the label text is a single attribute in the
underlying vector data. A configuration like this can be added to the style section of each vector based layer configuration:

```json
        "label": {
          "attribute": "name",
          "minResolution": 4.0,
          "outlineColor": "white",
          "outlineWidth": 2,
          "fillColor": "blue",
          "offsetX": 0,
          "offsetY": 15,
          "align": "center"
        }
```
Result:
image

Besides `attribute`, `outlineColor`, `outlineWidth`, `fillColor`, `minResolution`, `maxResolution` every  config property of [ol/style/Text](https://openlayers.org/en/latest/apidoc/module-ol_style_Text-Text.html) can be set, e.g. `offsetX` and `offsetY`.

More sophisticated templating can be discussed in #157.

### DIFF
--- a/src/factory/OlStyle.js
+++ b/src/factory/OlStyle.js
@@ -1,4 +1,6 @@
-import { Circle as CircleStyle, Icon as IconStyle, Fill, Stroke, Style } from 'ol/style';
+import {
+  Circle as CircleStyle, Icon as IconStyle, Fill, Stroke, Style, Text }
+  from 'ol/style';
 
 /**
  * Factory, which creates OpenLayers style instances according to a given config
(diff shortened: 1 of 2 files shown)


src/factory/OlStyle.js
```diff
@@ -109,6 +120,56 @@ export const OlStyleFactory = {
...
+  },
+
+  /**
+   * Creates and returns a basic OL style object for texts.
+   *
+   * @param {Object} labelConf Style config object for labels
+   * @returns {ol/style/Text} The OL style object for texts
+   */
+  getTextStyle (labelConf) {
+    const textConf = { ...labelConf };
```
JakobMiksch, line 132: why is this necessary? 
a comment might be helpful

src/factory/OlStyle.js
```diff
@@ -109,6 +120,56 @@ export const OlStyleFactory = {
...
+   * @param {Object} labelStyleConf
+   * @returns {ol/style/Style~StyleFunction}
+   *    Style function returning the OL style object enriched by texts / labels
+   */
+  getTextStyleFunction (style, labelStyleConf) {
+    const labelStyle = OlStyleFactory.getTextStyle(labelStyleConf);
+    const labelAttr = labelStyleConf.attribute;
+    return (feature, resolution) => {
+      // detect min/max resolution to show labels
+      // if nothing is confured labels are shown regardless of resolution
```
JakobMiksch, line 159: `confured` - what is meant here?

JakobMiksch (commented):
@chrismayer I tried it also with another layer in it works fine. 
Just two minor remarks regarding comments

src/factory/OlStyle.js
chrismayer: It ensures a clone of the config is created, so no in place modification of the config is done. Good point, I'll add a comment.

src/factory/OlStyle.js
chrismayer: `configured` , I'll address that.

src/factory/OlStyle.js
chrismayer: See e32d245

JakobMiksch:
Looks good. Ready to merge.

chrismayer:
Thanks again @JakobMiksch

PR #159, opened 2020-12-21
Remove "app" directory

JakobMiksch:
For custom applications it is useful to use Wegue as git submodule and symlink the `app` directory into Wegue. This is for example done here: https://github.com/meggsimum/wegue-demos This works in general, but the upgrading Wegue to a new commit is a bit cumbersome, because the template `app` directory overwrites the `app`symlink. Moreover git assumes that the Wegue git submodule is "dirty". 

This pull request removes the `app` directory, in order to make Wegue upgrade easier in custom applications.

I already tested `npm run init:app` and it works without problem. 

Things to check:
- [ ] make sure this works with Docker

TODO:
- [ ] keep README.md


chrismayer:
Thanks for this @JakobMiksch. I have to check in detail and if the Docker image is still working.

One first question: Is there a reason why you delete `app-starter/README.md` ?

JakobMiksch:
I deleted `app-starter/README.md` because it would be copied to `app` as well, which might be confusing. Otherwise we can modify the `init-app.js` script to exlude the `README.md`

chrismayer:
I'd vote for modification of the `init-app.js` script to exclude the `README.md` from copying.

justb4:
Thinking with you: I see the need for easier use of Wegue in applications. Up to now a fork/clone of the Wegue repo was assumed, but that is not very flexible, e.g. when having multiple app-repos. I think in the end we should aim for Wegue as an NPM module with a Wegue CLI "wegue create app" etc . Maybe this is easier than we think: e.g. https://zellwk.com/blog/publish-to-npm/.

Wegue as submodule can be useful as well. Not sure if using a symlink for `app` is a flexible way to go. Thinking of alternatives for using submodule...Also https://github.com/meggsimum/wegue-demos/blob/master/create-wegue-app.sh assumes a bash-environment...(shouldn't line 71 be `ln -s app wegue/app`?) Would be nice if we could run something like `npm run init:app ../app` from within the Wegue submodule dir and let the JS script figure out. In first instance it is mainly the `imports` that need `wegue/src` i.s.o. `..src/` from the app-template. But I foresee issues with build and Docker in this case...How does the symlink play with GitHub? It is within the submodule, what happens on commit/push and if someone else clones the app repo?

JakobMiksch:
Hi @justb4. Thanks for you feedback! Having a npm module would indeed be great. I try to be pragmatic for the moment and try to use wegue as git submodule. This is an attempt to make upgrading exisiting wegue-based apps easier. Because ideally an upgrade to the latest commit should not be more than a `git pull` inside the submodule. 

But I agree, that there are still some open questions how this symlink-approach works in production. Currently I am moving an old Wegue-app with this method and try to get some experience. The first thing I noticed was that Wegue also indexes the `app` directory. Hence, if the `app` directory could be removed from the Wegue repo, my upgrade process would be much easier.

My current directory structure looks like below. The app directory is located outside of Wegue. That's why I think `ln -s app wegue/app` should be okay.

```
.
├── app
│   ├── components
│   ├── static
│   ├── WguAppTemplate.vue
│   └── WguApp.vue
└── wegue
    ├── app -> ../app
    ├── app-starter
    ├── build
    ├── ci
    ├── config
    ├── Dockerfile
    ├── docs
    ├── embedded.html
    ├── index.html
    ├── LICENSE.md
    ├── node_modules
    ├── package.json
    ├── package-lock.json
    ├── README.md
    ├── screenshots
    ├── src
    ├── test
    └── util

```

justb4:
@JakobMiksch Ok, let's aim for the submodule scenario (you still need to look at [line 71 of create-wegue-app.sh](https://github.com/meggsimum/wegue-demos/blob/master/create-wegue-app.sh#L71). 

Think leaving out the `app` dir could work but you only can have one app (symlink) per repo. Two things on submodule usage, (IMHO) best to try out:

* git submodules need to be maintained for versioning as well: they represent not just a repo but also a commit-level (hash), optionally (I think) a branch from the original repo. Just doing `git pull` within your submodule dir only makes your local copy uptodate with Wegue `master`. You also need to add/commit/push those changes to increase your submodule's commit level.
* thinking of that I think that Wegue (or your repo) should then have a `.gitignore` for Wegue's `app` dir. Otherwise it will try to update the original Wegue repo with the symlink (which will fail for most users).

JakobMiksch:
I noticed a bug with this symlink apporach: https://github.com/meggsimum/wegue-demos/issues/1
Currently I am not sure if this a systematic problem of symlinks

JakobMiksch:
closing, because it seems that meggsimum/wegue-demos#1 cannot be fixed easily

PR #164, opened 2020-12-23
Fix position of attribution

JakobMiksch:
the position of the attribution was too high

before:
image

after:
image


chrismayer:
Nice fix, thanks @JakobMiksch!

PR #82, opened 2019-07-13
add basic Projection (definition) support for Map

justb4:
This PR provides basic support to use other projections than the default Web Mercator. It has the following changes:

* dependencies in package.json: add `proj4`
* config: additional Map attributes for projection and projection definitions for `proj4`
* `Map.vue` handle projection setting (default EPSG:3857 WM) for Map 
* add additional `Proj4` definitions if specified in config
* example config for Dutch projection (EPSG:28992) WMS base Layer and overlay
* some test lines, but bit unclear how to really test

Was new to Wegue, so hope the changes are at the proper places. Hope that is the intention, at least within NL most apps use our local projection. I realize that introducing projections also introduces complexity, don't know if that is the intention for Wegue. 

Thanks for the good work! Hopefully Wegue can be a follow-up for GeoExt!

PR #170, opened 2021-02-11
Fix default map projection (use projection of OpenLayers)

### LINKED ISSUE #166: TileGrid shifts XYZ-Basemaps about 5km
It seems like the TileGrid change causes a shift of XYZ-Basemaps (~5km). The "standard" OSM Basemap via the custom OL-source is not affected. See GIF below.  Below also the layer definition in  `app-config.json`
When I remove the `tileGrid` Property in the `LayerFactory`, everything works as expected.
@justb4 

```javascript
  /**
   * Returns an XYZ based tile layer instance due to given config.
   *
   * @param  {Object} lConf  Layer config object
   * @return {ol.layer.Tile} OL XYZ layer instance
   */
  createXyzLayer (lConf) {
    const xyzLayer = new TileLayer({
      name: lConf.name,
      lid: lConf.lid,
      displayInLayerList: lConf.displayInLayerList,
      visible: lConf.visible,
      opacity: lConf.opacity,
      source: new XyzSource({
        url: lConf.url,
        attributions: lConf.attributions,
      //  tileGrid: lConf.tileGrid,  // <-- this probably causes the shift
        projection: lConf.projection
      })
    });
```

Peek 2021-01-28 21-01

Layer Definition of `app-config.json`:
```json
    {
      "type": "VECTOR",
      "lid": "gewerbe",
      "name": "Gewerbe Mutterstadt",
      "url": "./static/data/gewerbe.geojson",
      "format": "GeoJSON",
      "visible": true,
      "selectable": true,
      "style": {
        "radius": 4,
        "strokeColor": "purple",
        "strokeWidth": 2,
        "fillColor": "rgba(155,153,51,0.5)",
        "label": {
          "attribute": "title",
          "minResolution": 4.0,
          "outlineColor": "white",
          "outlineWidth": 2,
          "fillColor": "black",
          "offsetX": 0,
          "offsetY": 15,
          "align": "center"
        }
      }
    },
    {
      "type": "XYZ",
      "name": "Carto Positron",
      "url": "http://basemaps.cartocdn.com/light_all/{z}/{x}/{y}.png",
      "lid": "positron",
      "displayInLayerList": true,
      "visible": true,
      "attributions": "© <a href='https://carto.com/attribution/' target='_blank'>CARTO</a> | <a href='https://openstreetmap.org/copyright' target='_blank'>© OpenStreetMap</a>-contributors | "
    }
```

chrismayer:
By introducing custom projections in Wegue (#82) , we set EPSG:3857 as default SRS of the map if nothing special has been defined. Therefore the projection has been declared in the code as `{code: 'EPSG:3857', units: 'm'}`, which differs from the projection definition of OpenLayers. This leads to an unwanted map shift in some edge cases (see #166).
  
This PR ensures that the definition of the default map projection ('EPSG:3857') - for the case that no special projection is defined
in the app-config - is derived by OpenLayers instead of defining it in the code.

Fixes #166

### DIFF
--- a/src/components/ol/Map.vue
+++ b/src/components/ol/Map.vue
@@ -16,6 +16,7 @@ import RotateControl from 'ol/control/Rotate';
 import Projection from 'ol/proj/Projection';
 import TileGrid from 'ol/tilegrid/TileGrid';
 import {register as olproj4} from 'ol/proj/proj4';
+import {get as getProj} from 'ol/proj';
 import Overlay from 'ol/Overlay';
 import {GPX, GeoJSON, IGC, KML, TopoJSON} from 'ol/format';
 import {Vector as VectorLayer} from 'ol/layer';
@@ -109,12 +110,13 @@ export default {
       olproj4(proj4);
     }
 
-    // Projection for Map, default is Web Mercator
+    // Projection for map, default is Web Mercator
+    let projection;
     if (!this.projection) {
-      this.projection = {code: 'EPSG:3857', units: 'm'}
+      projection = getProj('EPSG:3857');
+    } else {
+      projection = new Projection(this.projection);
     }
-
-    const projection = new Projection(this.projection);
 
     // Optional TileGrid definitions by name, for ref in Layers
     Object.keys(this.tileGridDefs).map(name => {


src/components/ol/Map.vue
```diff
@@ -109,13 +110,14 @@ export default {
       olproj4(proj4);
     }
 
-    // Projection for Map, default is Web Mercator
+    // Projection for map, default is Web Mercator
+    let projection;
     if (!this.projection) {
```
JakobMiksch, line 115: spontaneously I would turn around the if/else - e.g. `if (this.projection){ ... } else {}` 
I find  this more  readable. But that is pure eye-candy  ...

JakobMiksch:
@chrismayer good catch finding the cause of the problem! the code looks fine and works for me. Only a minor non-blocking comment. Otherwise ready to merge.

JakobMiksch (approved):

Issue #141, opened 2020-09-03
Make online application demo available again


chrismayer:
Due to the fact that we deliver the new Wegue website on `gh-pages` (https://meggsimum.github.io/wegue, see #136 ), the online application demo (which was deployed there before) is not available anymore. It should be restored somewhere in our project's `gh-pages`.

justb4:
Hmm, a bit a puzzle, also in the light of CNAME (deleted?) and a custom domain. 
Somehow we need to have docs and the demo under the same dir tree. Thinking loud I see two options:

* use the `gh-pages` branch for the website, on post-commit copy the `docs` tree to there and build/copy the demo there
* use `master` and on post-commit build/copy the demo app to the `docs` tree

In all cases I would prefer maintaining the docs on the `master` branch and the second option above. 
Lately I started using GitHub Actions over Travis, this can help with the above.

chrismayer:
I'd also prefer the "use master and on post-commit build/copy the demo app to the docs tree" solution (if possible). I have no strong opinion whether we use Travis or github actions. Whatever does the trick is fine for me.

Issue #140, opened 2020-09-03
website updates


justb4:
The website can use some more updates:

- [ ] the Wegue demo app is missing, link on cover page refers to site itself
- [ ] a domain name coupled to CNAME would be nice
- [ ] Docker setup section

(Fill in other items here)

JakobMiksch:
party done with #171

PR #171, opened 2021-02-16
Docs: link to a demo application

JakobMiksch:
image

Adds a link to the demo button: https://apps.meggsimum.de/wegue-demos/global/

We can of course also take another demo app for the link

fixes: #141
also partly: #140


chrismayer (approved):
OK for the moment as intermediate solution. In the mid-term we should deploy the offical example apps somewhere (as stated in #141)

PR #173, opened 2021-02-17
Make helpwin configurable

### LINKED ISSUE #168: Make help window content configurable
The content of the `HelpWin` can currently only be set / changed by changing the source code / overwriting the component.

It would be very useful to have a configurable way of setting the content (preferably HTML) like we can do for other texts like in the footer.

JakobMiksch:
Makes it possible to control the content of the help window using configuration.

fixes #168

### DIFF
--- a/src/components/helpwin/HelpWin.vue
+++ b/src/components/helpwin/HelpWin.vue
@@ -1,25 +1,28 @@
 <template>
 <!--remove <v-toolbar-side> -->
   <v-card class="wgu-helpwin">
-    <v-toolbar :color="color" class="" dark>
+    <v-toolbar :color="color" dark>
       <v-icon>{{ icon }}</v-icon>
-      <v-toolbar-title>{{ title }}</v-toolbar-title>
+      <v-toolbar-title v-if="title">{{ title }}</v-toolbar-title>
       <v-spacer></v-spacer>
       <v-icon @click="onWinXClose">close</v-icon>
     </v-toolbar>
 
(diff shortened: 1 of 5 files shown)


src/components/helpwin/HelpWin.vue
```diff
@@ -29,18 +32,26 @@
...
-      content: {type: String, required: false, default: '<h3>WebGIS with OpenLayers and Vue.js</h3> Template and re-usable components for webmapping applications with OpenLayers and Vue.js'},
-      infoLink: {type: String, required: false, default: 'https://github.com/meggsimum/wegue'},
-      infoLinkText: {type: String, required: false, default: 'More info'}
+      icon: {type: String, required: false, default: 'help'}
     },
     data () {
       return {
-        show: false
+        show: false,
+        title: '',
```
chrismayer, line 40: I think the apply of the config data can be done directly `title: this.$appConfig.modules['wgu-helpwin'].windowTitle`, so implementing the `mounted` function could be obsolete.

JakobMiksch: that's right. 
I added directly assigned the configs in the  `data` property

chrismayer:
`app-starter/app-conf-projected.json` and `app-starter/app-conf-minimal.json` should also get the settings for the help window.

JakobMiksch:
@chrismayer I added the helpwin to `minimal` and `projected` as well.

chrismayer (approved):
LGTM, thanks!

PR #174, opened 2021-02-17
Update URL of demo WFS layer "gas stations"

JakobMiksch:
The URL of of the demo WFS layer "gas stations" has changed.


chrismayer (approved):
Thanks, LGTM!

PR #190, opened 2021-03-25
Enable hoverable configuration for WFS layers

### LINKED ISSUE #162: Enable "hoverable" configuration for WFS layers
For vector layers and vectortile layers we support a `hoverable` configuration, which allows to enable feature hovering on the map. On mouseover a little popup showing an attribute is displayed on the map. For WFS it should also be enabled / implemented.

Compare 
https://github.com/meggsimum/wegue/blob/master/src/factory/Layer.js#L259 
https://github.com/meggsimum/wegue/blob/master/src/factory/Layer.js#L287

After implementation the property `hoverable` should also be described in the WFS layer docs.

chrismayer:
Enables `hoverable` and `hoverAttribute` configuration for WFS layers. Also adds the corresponding docs.

Closes #162


JakobMiksch:
@chrismayer looks fine. Ready to merge.

chrismayer:
Thanks for the review!

PR #191, opened 2021-03-25
Several minor vector hover enhancements

chrismayer:
- Document `hoverable` + `hoverAttribute` in all necessary places in docs
  - Remove fixed with for vector hover tooltip element, using a max-width of 200px instead
  - Enable WFS layer hovering to `projected` example app

Should be merged after #190 is merged.


JakobMiksch:
@chrismayer I tested it and it works well. Code looks good as well. Please merge.

JakobMiksch (approved):

chrismayer:
Thanks for the review, I'll merge ...

PR #195, opened 2021-03-26
Fix feature selection with layer / feature hovering enabled

### LINKED ISSUE #193: Feature label prevents selection
The updated feature label https://github.com/meggsimum/wegue/commit/c922a04fe8b0e3ab0c232984ecfa2385dd3e4f5c prevents selecting a feature by clicking on it.

If I remove the `hoverAttribute` config in the appConfig, the selection works like expected.

Peek 2021-03-26 11-30

chrismayer:
This fixes the broken selection of a feature when the feature hovering is enabled. This error was introduced with #191.

Fixes #193

### DIFF
--- a/src/components/ol/Map.vue
+++ b/src/components/ol/Map.vue
@@ -266,6 +266,8 @@ export default {
       // wrap the tooltip span in a OL overlay and add it to map
       me.overlay = new Overlay({
         element: overlayEl,
+        stopEvent: false,
+        className: 'wgu-hover-ol-overlay',
         autoPan: true,
         autoPanAnimation: {
           duration: 250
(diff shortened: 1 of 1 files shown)


src/components/ol/Map.vue
```diff
@@ -266,6 +266,8 @@ export default {
       // wrap the tooltip span in a OL overlay and add it to map
       me.overlay = new Overlay({
         element: overlayEl,
+        stopEvent: false,
+        className: 'wgu-hover-ol-overlay',
```
JakobMiksch, line 270: why is `classname` necessary?
I cannot find a reference to it in the code base.

JakobMiksch (approved):
@chrismayer I tried it and now the selection works like expected. Just  one non-blocking question inline.

src/components/ol/Map.vue
chrismayer: See section `stopEvent` in https://openlayers.org/en/latest/apidoc/module-ol_Overlay-Overlay.html

PR #196, opened 2021-03-26
Ignore measure-layer for permalink

### LINKED ISSUE #194: Permalink shows 'measure-layer'
`http://localhost:8081/#z=6.9262&c=9.6097%2C44.5568&r=0&l=osm-bg%2Cearthquakes%2Cgas-wfs%2CShops%2Cmeasure-layer`

Since https://github.com/meggsimum/wegue/commit/3bed90cd3e70e34b8b1ff5bf2a1e7b6f1f4fc3de the "Measure Layer" has a `lid` (see the last item of the URL above).

This causes the permalink to also show the "Measure Layer" which is probably not desired. 

The fix should probably be done in [PermalinkController.js](https://github.com/meggsimum/wegue/blob/43b825c5f0acd78a6e229eda4da1fda87bf4986c/src/components/ol/PermalinkController.js)

chrismayer:
This ensures that the layer for the measurement geometries (`measure-layer`) is not listed in the permalink.

Fixes #194

### DIFF
--- a/src/components/ol/PermalinkController.js
+++ b/src/components/ol/PermalinkController.js
@@ -13,6 +13,7 @@ export default class PermalinkController {
   projection = null;
   conf = null;
   urlParams = null;
+  ignoreLayers = ['measure-layer'];
 
   constructor (map, permalinkConf) {
     this.map = map;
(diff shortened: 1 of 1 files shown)


JakobMiksch (approved):
@chrismayer I tried it and it works fine. Code looks good as well.

chrismayer:
Thanks for the review!

PR #197, opened 2021-03-26
AttributeTable-Map interaction

JakobMiksch:
This PR creates an interaction between the AttributeTable and its corresponding  layer:

- a click on a row zooms to the feature in the map
- a click on a row selects a feature in the map, if this layer is `selectable`
- selecting a feature on the map shows the corresponding row in the table, if it is currently opened
- the interaction between the layer and the map can be configured with `activateTableMapInteraction`
- ~~the AttributeTable-Map interaction is only enabled on desktop views, because it is not applicable on mobile views~~ it works for mobile as well
- the loading process of the features is refactored

## TODO

- [x] Re-Use `MapInteraction.js` util. If possible move code there. see https://github.com/meggsimum/wegue/pull/201
- [x] Tests are missing. 
    - There is the need to mock `$vuetify` (see [here](https://vuetifyjs.com/en/getting-started/unit-testing/#mocking-vuetify) or [GithHub Issue](https://github.com/vuetifyjs/vuetify/issues/11278))
    - The property `activateTableMapInteraction` is somehow not recognized yet

referencing issue https://github.com/meggsimum/wegue/issues/192

Peek 2021-03-26 18-16

### DIFF
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -242,7 +242,8 @@
     },
     "wgu-attributetable": {
       "target": "menu",
-      "win": true
+      "win": true,
+      "activateTableMapInteraction": true
     }
   }
 }
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -243,7 +243,7 @@
     "wgu-attributetable": {
       "target": "menu",
       "win": true,
-      "activateTableMapInteraction": true
+      "syncTableMapSelection": true
     }
   }
 }
--- a/app/README.md
+++ b/app/README.md
@@ -1,12 +1,5 @@
-# Wegue App Dir
-This directory will contain your Wegue Application.
-By using the command `npm run init:app` this directory is populated with
-the contents from the `app-starter/` dir.
+# Wegue App Starter Dir
+This directory contains a Wegue starter app.
+By using the command `npm run init:app` this directory is copied to the `app/` directory.
+For now this copy will not overwrite files already present in the `app/` dir.
 
-You can develop your Wegue app by modifying/adding files here, under the `app/` dir, for example:
-
-* modifying the core `.vue` files like `app/WguApp.vue` and `app/WguAppTemplate.vue`
-* placing or modifying custom/app-specific components under `app/components`
-* placing or modifying  static content like configuration, data, icons and css under `app/static`.
-
-The default config file is `app/static/app-conf.json`.
--- a/app/README.md
+++ b/app/README.md
@@ -1,5 +1,12 @@
-# Wegue App Starter Dir
-This directory contains a Wegue starter app.
-By using the command `npm run init:app` this directory is copied to the `app/` directory.
-For now this copy will not overwrite files already present in the `app/` dir.
+# Wegue App Dir
+This directory will contain your Wegue Application.
+By using the command `npm run init:app` this directory is populated with
+the contents from the `app-starter/` dir.
 
+You can develop your Wegue app by modifying/adding files here, under the `app/` dir, for example:
+
+* modifying the core `.vue` files like `app/WguApp.vue` and `app/WguAppTemplate.vue`
+* placing or modifying custom/app-specific components under `app/components`
+* placing or modifying  static content like configuration, data, icons and css under `app/static`.
+
+The default config file is `app/static/app-conf.json`.
--- a/docs/module-configuration.md
+++ b/docs/module-configuration.md
@@ -91,4 +91,4 @@ Module identifier: `wgu-attributetable`
 
 | Property                    | Meaning   | Example |
 |-----------------------------|:---------:|---------|
-| activateTableMapInteraction | Clicking on a row zooms to the respective feature. If the layer is `selectable` the feature will also be selected. Selecting a feature on the map selects the corresponsing row in the table. | `"activateTableMapInteraction": true` |
+| syncTableMapSelection | Clicking on a row zooms to the respective feature. If the layer is `selectable` the feature will also be selected. Selecting a feature on the map selects the corresponsing row in the table. | `"syncTableMapSelection": true` |
--- a/docs/module-configuration.md
+++ b/docs/module-configuration.md
@@ -91,4 +91,4 @@ Module identifier: `wgu-attributetable`
 
 | Property                    | Meaning   | Example |
 |-----------------------------|:---------:|---------|
-| activateTableMapInteraction | Works only for desktop view, it is automatically disabled on mobile view. Clicking on a row zooms to the respective feature. If the layer is `selectable` the feature will also be selected. Selecting a feature on the map selects the corresponsing row in the table. | `"activateTableMapInteraction": true` |
+| activateTableMapInteraction | Clicking on a row zooms to the respective feature. If the layer is `selectable` the feature will also be selected. Selecting a feature on the map selects the corresponsing row in the table. | `"activateTableMapInteraction": true` |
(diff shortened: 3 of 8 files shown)


JakobMiksch:
@chrismayer I finally made the tests running. This is the relevant part in the code:

```js
   vuetify = new Vuetify()

      comp = shallowMount(AttributeTable, {
        mocks: {
          $vuetify: {
            breakpoint: {
              t: (val) => val
            }
          }
        },
        vuetify
      });
```
I got this snippet from this [Issue](https://github.com/vuetifyjs/vuetify/issues/11278). Apparently this was a problem for more people.

app-starter/static/app-conf.json
```diff
@@ -229,7 +229,8 @@
     },
     "wgu-attributetable": {
       "target": "menu",
-      "win": true
+      "win": true,
+      "activateTableMapInteraction": true
```
chrismayer, line 233: I'd prefer something like

```suggestion
      "syncTableMapSelection": true
```

app/README.md
```diff
@@ -1,12 +1,5 @@
-# Wegue App Dir
-This directory will contain your Wegue Application.
-By using the command `npm run init:app` this directory is populated with
-the contents from the `app-starter/` dir.
+# Wegue App Starter Dir
```
chrismayer, line 1: I think this is somehow unrelated and even wrong since you are overiting the README in the `app/` folder with the content of `app-starter`, right?

docs/module-configuration.md
```diff
@@ -89,4 +89,6 @@ The attribute table displays features of vector layers. It is possible to specif
 
 Module identifier: `wgu-attributetable`
 
-No additional config options besides the general ones.
+| Property                    | Meaning   | Example |
+|-----------------------------|:---------:|---------|
+| activateTableMapInteraction | Clicking on a row zooms to the respective feature. If the layer is `selectable` the feature will also be selected. Selecting a feature on the map selects the corresponsing row in the table. | `"activateTableMapInteraction": true` |
```
chrismayer, line 94: See comment above regarding naming of the property

src/components/attributeTable/AttributeTable.vue
```diff
@@ -15,19 +15,52 @@
...
+    uniqueRecordKeyName: {type: String, required: false, default: 'fid'},
+
+    /**
+     * How many rows a page of the table should have.
+     * Should be manually adjusted with tableHeight.
+     */
+    rowsPerPage: {type: Number, required: false, default: 10},
+
+    /**
+     * The hight of the table in pixel.
```
chrismayer, line 53:
```suggestion
     * The height of the table in pixel.
```

src/components/attributeTable/AttributeTable.vue
```diff
@@ -15,19 +15,52 @@
...
+     * Should be manually adjusted with rowsPerPage.
+     */
+    tableHeight: {type: Number, required: false, default: 272},
+
+    /** If map and table should be synced */
+    activateTableMapInteraction: {type: Boolean, required: false, default: true},
```
chrismayer, line 59: See comment above regarding naming of the property

src/components/attributeTable/AttributeTable.vue
```diff
@@ -41,21 +74,194 @@ export default {
...
   watch: {
     layerId () {
       this.populateTable()
+    },
+    features () {
+      this.records = this.features.map(
+        feature => {
+          const record = feature.getProperties();
+          // set feature id
+          record[this.uniqueRecordKeyName] = feature.getId()
```
chrismayer, line 106: Does feature.getId() always return a valid value? Or can it be undefined / null at some time? Please check https://openlayers.org/en/latest/apidoc/module-ol_Feature-Feature.html#getId especially the part with "remote source".

chrismayer (changes requested):
Very nice addition to the AttributeTable. Thanks for this!

I'd added some comments in the code, which should be addressed or at least checked. 

Furthermore I am facing the following when testing the feature in the demo app: When I select a feature (either clicking a row in the AttributeTable or by selecting a map feature) the table row is optically not highlighted. Is this wanted behavior? Doers Vuetify not support this?

image

app/README.md
JakobMiksch: Right. I reverted it with 8811c22a12549cda8e88b8ab16e65547c60e0ae0

JakobMiksch:
@chrismayer regarding the issue with the highlighting of the row.

Vuetify actually does highlight the row, but the color is very pale, and not that easy to see.  This is the default setting. This is of course possible to change, but it requires some CSS - [Blog Post](https://www.javaer101.com/en/article/49856669.html) - [Example](https://codepen.io/huntleth/pen/eYOrWog)

src/components/attributeTable/AttributeTable.vue
JakobMiksch: done with 73d37cc

docs/module-configuration.md
JakobMiksch: done with 73d37cc

src/components/attributeTable/AttributeTable.vue
JakobMiksch: @chrismayer very good observation!
In fact the tested layers do all return valid values from `feature.getId()`. Not only WFS, but also GeoJSON and KML. Apparently all of them are considered as "remote source". 
Getting IDs might be a problem for not-"remote sources" like for layers where features are created manually. However, Currently I do not see a scenario where this happens. We should for sure note this in an issue.

How shall we proceed with this? Fixing or keeping it in mind?

chrismayer: Not sure here. Should definitely be tackled but maybe not here. Maybe setting a random ID by setId for non-remote sources would be an option. But this should be done at layer source creation or similar. I created an issue for that #210.

chrismayer:
> @chrismayer regarding the issue with the highlighting of the row.
> 
> Vuetify actually does highlight the row, but the color is very pale, and not that easy to see. This is the default setting. This is of course possible to change, but it requires some CSS - [Blog Post](https://www.javaer101.com/en/article/49856669.html) - [Example](https://codepen.io/huntleth/pen/eYOrWog)

OK, that is fine for me. We could add a convenient way to change the color in the future. Maybe in combination with the issue #202.

JakobMiksch:
@chrismayer I rebased this PR to the latest `master` 
Let me know if you have more remarks. Otherwise it should be ready to merge. Squashing the commits is also fine.

chrismayer:
Does this already use the changes from #201 ? A stated in the open TODO `Re-Use MapInteraction.js util. If possible move code there`

JakobMiksch:
@chrismayer thanks for pointing me to the TODO item. I moved functionality to the new `MapInteraction.js` util with e15e1ed.

chrismayer (approved):
LGTM, thnaks for your ongoing effort. I'll merge now.

fschmenger:
Regarding the specific resize and OL refresh workarounds in AttributeTableWin.vue imo we should rather try and fix this on the AppTemplate level, e.g. provide a dedicated “application bottom container” (if the attribute table window is going to stay there).

geomer GmbH
Felix Schmenger
Entwicklung
Im Breitspiel 11 B
69126 Heidelberg
Germany

Tel.:+49 (6221) 89458-49
Fax: +49 (6221) 89458-79
E-Mail: ***@***.******@***.***>
URL: www.geomer.de<https://www.geomer.de/>

[partner]
[abobutton]<https://www.geomer.de/anmeldung-newsletter>

Firmensitz: Im Breitspiel 11 b, 69126 Heidelberg, Deutschland
Registergericht: Amtsgericht Mannheim, HRB 336266
Geschäftsführer: Martin Schroeder

JakobMiksch:
@fschmenger thanks for the suggestion. Would be great to find a more elegant solution to resizing of the map.

PR #201, opened 2021-04-22
Refactor SelectInteraction

JakobMiksch:
- Move the creation of the selectInteraction to an util class. 
- Extent style append function and move it a separate util 
- this makes it possible to combine the select style with the current style

- [x] TODO: selection-style - it should both be possible to replace the original style and to combine the styles

Peek 2021-04-22 11-33

### DIFF
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -64,7 +64,14 @@
         "name": "Name",
         "email": "Email",
         "website": "Website"
-      }
+      },
+      "selectStyle": {
+        "radius": 10,
+        "strokeColor": "gray",
+        "strokeWidth": 5,
+        "fillColor": "rgb(255, 255, 0, 0.2)"
+      },
+      "doAppendSelectStyle": true
     },
     {
       "type": "WFS",
@@ -88,7 +95,13 @@
       "attributions": "© <a href=\"https://www.openstreetmap.org/copyright\">OpenStreetMap</a> contributors.",
       "columnMapping": {
         "name": "Name"
-      }
+      },
+      "selectStyle": {
+        "textIcon": "star",
+        "font": "normal 30px Material Icons",
+        "fillColor": "black"
+      },
+      "doAppendSelectStyle": false
     },
 
     {
--- a/docs/map-layer-configuration.md
+++ b/docs/map-layer-configuration.md
@@ -25,34 +25,36 @@ The following properties can be applied to all map layer types
 
 ## VECTOR
 
-| Property           | Meaning | Example |
-|--------------------|:---------:|---------|
-| **type**           | Indicator that the layer is a vector layer, always `VECTOR` here  | `"type": "VECTOR"` |
-| **url**            | The URL to the vector data resource (file) | `"url": "./static/data/2012_Earthquakes_Mag5.kml"` |
-| **format**         | The format of the data linked in `url` (either `KML` or `GeoJSON` ) | `"format": "KML"` |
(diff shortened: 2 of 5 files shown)


JakobMiksch:
@chrismayer I made the appending of the selectStyle optional and default to `false`. The example below shows both options:
- the select style of the shops are appended with a circle
- the select style of the gas stations are replaced by a black star

the respective property in the layer config is `doAppendSelectStyle`. If not set, the select style replaces the original style.

Peek 2021-04-23 09-25

app-starter/static/app-conf.json
```diff
@@ -64,7 +64,14 @@
...
         "website": "Website"
-      }
+      },
+      "selectStyle": {
+        "radius": 10,
+        "strokeColor": "gray",
+        "strokeWidth": 5,
+        "fillColor": "rgb(255, 255, 0, 0.2)"
+      },
+      "doAppendSelectStyle": true
```
chrismayer, line 74: Maybe better `appendSelectStyle`

app-starter/static/app-conf.json
```diff
@@ -88,7 +95,13 @@
...
       "columnMapping": {
         "name": "Name"
-      }
+      },
+      "selectStyle": {
+        "textIcon": "star",
+        "font": "normal 30px Material Icons",
+        "fillColor": "black"
+      },
+      "doAppendSelectStyle": false
```
chrismayer, line 104: Maybe better `appendSelectStyle`

docs/map-layer-configuration.md
```diff
@@ -25,34 +25,36 @@ The following properties can be applied to all map layer types
...
+| **url**             | The URL to the vector data resource (file) | `"url": "./static/data/2012_Earthquakes_Mag5.kml"` |
+| **format**          | The format of the data linked in `url` (either `KML` or `GeoJSON` ) | `"format": "KML"` |
+| selectable          | Boolean value, whether the features of the layer can be selected by click in order to display the attributes in a window | `"selectable": true` |
+| hoverable           | Boolean value, whether the features of the layer can be hovered in order to display an attribute (see `hoverAttribute`) in a tooltip  | `"hoverable": true` |
+| hoverAttribute      | Attribute to be shown if a feature of the layer is hovered. Only has an effect if `hoverable` is set to `true`  | `"hoverAttribute": "name"` |
+| style               | Object to define a rendering style for the features of the layer  | see [style](map-layer-configuration?id=style-for-vectorlayers) |
+| selectStyle         | The style for a selected feature | see [style](map-layer-configuration?id=style-for-vectorlayers) |
+| doAppendSelectStyle | If the selectStyle should be appended to the original style, defaults to `false` | `"doAppendSelectStyle": true` |
```
chrismayer, line 38: Maybe better `appendSelectStyle`

chrismayer (approved):
LGTM; thnaks for this.

app-starter/static/app-conf.json
JakobMiksch: @chrismayer I intentionally called it `doAppendSelectStyle` to make clear that the value should be a Boolean and not a style definition object. But we can also rename it.

JakobMiksch: Just let me know your preference than I can change it.

chrismayer: Right, so it is clearer. Keep it as it is.

JakobMiksch:
@chrismayer thanks for your review. If you do not have any other remarks, this should be ready to merge.

chrismayer:
Fine, I'll merge now. Thanks again.

PR #212, opened 2021-05-04
Material icons

fschmenger:
This branch adds material design icons (@mdi/font) and material icon (material-icons) NPM packages. While the first isn't strictly required, it gives us a bunch of new icon options. The difference in package sizes seem to be moderate - biggest change is in source maps.

Head
static/css/app.f0d6ecc014b266ef49f5ecf30858ce65.css     455 kB       1  [emitted]  [big]  app      
static/css/app.f0d6ecc014b266ef49f5ecf30858ce65.css.map     592 kB       1  [emitted]         app 

90140d3b6e6ab155d34dbefe7d4e1344272ccbe2: Added @mdi/font
static/css/app.edf256455dfa66015e7356a48407b18d.css     723 kB       1  [emitted]  [big]  app
static/css/app.edf256455dfa66015e7356a48407b18d.css.map    1.11 MB       1  [emitted]         app

43daa29b649da8f5648f1e42c274ba6a3ca2c197: Added material-icons
static/css/app.f246aeeafd13987c68f24912f85c6deb.css     725 kB       1  [emitted]  [big]  app
static/css/app.f246aeeafd13987c68f24912f85c6deb.css.map    1.12 MB       1  [emitted]         app

PR #213, opened 2021-05-11
Application sidebar

fschmenger:
This feature branch adds an application wide sidebar to the project template, which serves as a container for module content. The sidebar is only visible, if at least one module has been configured to use the sidebar target.

Wegue_sidebars

In app.conf.json, the "win" property of "modules" has been changed to type string accepting the values {floating, sidebar}, e.g. 

`"modules": {
    "wgu-layerlist": {
      "target": "menu",
      "win": "sidebar"
    },
    "wgu-measuretool": {
      "target": "menu",
      "win": "floating"   
    }
 }`
A boolean value is still accepted for backward compatibility, 

Additionally to make sidebar windows look a little nicer, the config accepts 2 new module properties, which can be applied to floating windows too:
* 'minimizable': Option to minimize windows and show a minimizer next to the window closer (default is true).
* 'backgroundImage': Optional background image for the window header.

### DIFF
--- a/app-starter/WguAppTemplate.vue
+++ b/app-starter/WguAppTemplate.vue
@@ -20,6 +20,7 @@
             :is="moduleWin.type" :key="index" :ref="moduleWin.type"
             :win="moduleWin.win" :color="baseColor"
             :backgroundImage="moduleWin.backgroundImage"
+            :minimizable="moduleWin.minimizable"
           />
       </template>
     </wgu-app-sidebar>      
@@ -45,6 +46,7 @@
         :draggable="moduleWin.draggable"
         :initPos="moduleWin.initPos"
         :backgroundImage="moduleWin.backgroundImage"
+        :minimizable="moduleWin.minimizable"
       />
     </template>
 
@@ -147,7 +149,8 @@
               win: moduleOpts.win,
               draggable: moduleOpts.draggable,
               initPos: moduleOpts.initPos,
-              backgroundImage: moduleOpts.backgroundImage
+              backgroundImage: moduleOpts.backgroundImage,
+              minimizable: moduleOpts.minimizable
             });
           }
         }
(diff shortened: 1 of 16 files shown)


fschmenger:
I missed, that the window minimizer uses an icon from the "material design icons" set, which is not contained in Wegue yet - and part of PR #212. Let me know how to proceed on this one. We could either
* merge #212 first to provide consistency
* or if you're planning on not to accept #212, I can try and look for an alternate icon.

chrismayer:
Very cool, this brings up a ton of new possibilities for layouting a Wegue app. Very much appreciated. Especially tahnks for having the backwards compatibility in the config. Makes things much easier. I think this is good to go.

Some questions / ideas:
  - It would be cool to have a module initially visible when it is rendered in the sidebar.
  - It would be cool to configure the initial visibility of the sidebar
  - An example app showing the sidebar feature would be very helpful. Either adapting the default example or the projected one or bringing in completely new example app

Surely nothing, what is blocking this PR. Could be tackled in the future.

fschmenger:
Yes, i like the ideas:

>     * It would be cool to have a module initially visible when it is rendered in the sidebar.
I agree. It looks a little empty on application start. My idea was to place the geocoder search bar there (google greats you) but that requires some work on the module.

>     * It would be cool to configure the initial visibility of the sidebar
* An initial sidebar visibility option can be achieved very easily, we just have to define the config option we want for that. Be aware though, that the behavior is different for small screens (sidebar becomes modal and is initially closed), so the option might not be feasible / have the desired effect.

>     * An example app showing the sidebar feature would be very helpful. Either adapting the default example or the projected one or bringing in completely new example app
* I can certainly add and additional config to the template, which sets all module 'win' options to 'sidebar'

I see you are already working on tracking some of the comments within issues. Good job! I was just about to propose the same.

chrismayer:
Thanks for your comments @fschmenger :+1: . I'll merge this and extract some more issues for sustainability.

Thanks again for your work!

Issue #202, opened 2021-04-29
Use Vuetify Theming


chrismayer:
It would be very handy to use the Vuetify theming mechanism. Goal would be to get rid of the `baseColor` property in the app-conf, which has to be passed through all components and instead a defining an optional Vuetify theme in the app-config.
Default style of a Wegue app could be a predefined standard Vuetify theme.

See 
  - https://v2.vuetifyjs.com/en/customization/presets/ 
  - https://vuetifyjs.com/en/features/presets/#generator 
  - https://material.io/design/material-theming/implementing-your-theme.html#color

JakobMiksch:
Hi @ntma , it seems there is no current work done on this topic. Hence, a contribution would be very welcome. 
Feel free to ask if there any questions. There is a also a Wegue chat on [gitter](https://gitter.im/wegue/community)

ntma:
Perfect, I'll see what I can do 👍

ntma:
Hi @fschmenger ,

Thank you for this outline :+1: 

Allow me to number your bullets to ease the discussion. So:
1. seems to be fairly easy. Wegue will start using the 'primary' and 'secondary' classes to theme the UI according to the material design system;
2. no questions here for now;
3. basically instead of providing a `baseColor`, we want to provide a preset config. Should this config be placed in the `app-config.json`?

fschmenger:
Hi @ntma, I haven't been very concise in my description but happy to see you got it right!

Yes, regarding point 3 I would suggest to support an optional declaration of a top level 'theme' attribute in app-config.json (this is what I meant when refering to 'application context'). 
I'm a bit busy until mid of next week but will follow this issue in case you have any further question.

PR #219, opened 2021-05-12
Module property initialization

fschmenger:
The purpose of this change-set is to forward the complete module configuration found in app-conf.json for each module. The configuration can then be picked up within each module via Vue props, rather than having to access the global $appConfig.modules.

The idea behind is:
* Provide better encapsulation of module configurations.
* Make it easier to validate properties (by using the implicit mechanism provided by Vue).
* Easily support the new properties introduced by the base Module cards - which now can all be configured via the config-
* Simplify the module initialization code.
* Revive some old properties like e.g. 'title', that we're not passed through from the app-config.json into the modules.

### DIFF
--- a/app-starter/WguAppTemplate.vue
+++ b/app-starter/WguAppTemplate.vue
@@ -17,10 +17,8 @@
     <wgu-app-sidebar v-if="sidebarWins.length">
         <template v-for="(moduleWin, index) in sidebarWins">
           <component
-            :is="moduleWin.type" :key="index" :ref="moduleWin.type"
-            :win="moduleWin.win" :color="baseColor"
-            :backgroundImage="moduleWin.backgroundImage"
-            :minimizable="moduleWin.minimizable"
+            :is="moduleWin.type" :key="index" :color="baseColor"
+            v-bind="moduleWin"
           />
       </template>
     </wgu-app-sidebar>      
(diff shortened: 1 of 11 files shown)


fschmenger:
This change only affects the properties declared in the 'modules' array of app-conf.json. 

* Some modules, e.g. Maps use global fields from app-conf.json to populate their configuration data.
* AppTemplate and AppHeader also access the global $appConfig object.

We could take the idea of this change-set a step further and consider to access the $appConfig only at top level program initialization. From there we could then rely on properties to pass the configuration through the component tree. IMO this would be nice, however requires a bit more work.

fschmenger:
Another little refactoring I propose would be to move the 'darkLayout' property from the individual modules and make it a global setting. Since the hamburger menu now shares the same background color as the toolbar and the window headers, there should be no drawbacks. This probably could be reviewed in conjunction with #202.

chrismayer (approved):
LGTM, thanks @fschmenger. This is an important step towards the encapsulation of modules as a pre-step for having a Wegue core library in the future.
Big thanks for also considering backwards compatibility again.

chrismayer:
Also here gain for transparency: There will be some more PRs tackling some refactoring tasks. The last step will be to adjust the unit tests at once to avoid duplicate work, so it is a know issue that this (and some follow ups) will break the test suite of the master branch for a couple days.

PR #220, opened 2021-05-20
Fix unit tests

fschmenger:
This adds new unit tests and fixes broken tests due to recent modifications of Wegue.
I also added missing dependencies like Vuetify to the unit tests to get rid of the warning bloat.


fschmenger:
Some further notes:

* Is there any purpose for 'my.karma.conf.js'? To me it looks obsolete so we might as well remove it.
* Someone has to go over the warnings for the FeatureInfoWindow tests, if we're going to revive the component.

chrismayer (approved):
This is really cool, thanks for also tackling this as finalization of your refactoring work @fschmenger.

justb4:
Great work @fschmenger ! Fixing JS/Vue unit tests, especially getting rid of warnings, is hard and repetitive work (at least for me)! 

 'my.karma.conf.js' was I think my attempt to run a single spec. Later adapted, see README in test dir:
*In Wegue testing we added the custom option `--spec-file`*, so yes: obsolete now.

fschmenger:
Thanks for the feedback @justb4 ! I created a PR to remove 'my.karma.conf.js'

Issue #203, opened 2021-04-29
Add background layer switcher


chrismayer:
Adding a separate tool / module to switch background layers. Detailed layout has to be done.

Inspiration by existing solutions:
  - https://www.geoportal.de/map.html?map=tk_01-stromnetz
  - https://geoportal.brandenburg.de/de/cms/portal/start

fschmenger:
I can start working on this feature. Before i get going:

Is any of the existing solutions mentioned above based on Wegue? Or is there otherwise any preexisiting code you would like me to incorporate?

Design-wise I would go into the direction of the examples you provided: A custom module will place a toggle button overlay on the map and then open a sort of sheet to pick one of the background layers.

chrismayer:
> I can start working on this feature.

Cool, very much appreciated.

> Is any of the existing solutions mentioned above based on Wegue? Or is there otherwise any preexisiting code you would like me to incorporate?

As far as I know there is no Wegue based solution or at least I am not aware of it.

> Design-wise I would go into the direction of the examples you provided: A custom module will place a toggle button overlay on the map and then open a sort of sheet to pick one of the background layers.

Yes, sound good :+1:

fschmenger:
I noticed, that the docs already define a flag 'isBaseLayer' for this purpose - which currently has no effect. I will use this as the marker for background layers.

fschmenger:
I`m almost there with the basic functionality. Now I'm thinking about the layer preview images, which should appear on the selection control:

* The simplest solution apparently is to provide image resources for each layer, although not very maintainable.
* We could include the ol-ext package, as it provides layer previews - [example here](https://viglino.github.io/ol-ext/examples/layer/map.preview.html)
* Another option would be to just incorporate the code from [https://github.com/Viglino/ol-ext/blob/master/src/layer/GetPreview.js]( https://github.com/Viglino/ol-ext/blob/master/src/layer/GetPreview.js) and tweak it (maybe go without modifications to the OL prototypes).

Personally I would prefer to create the previews implicitly without having to provide images but anything goes. An opinion on this is strongly appreciated!

chrismayer:
I'd also prefer to have an automatically generated preview images at least for layer types where this is easily doable TileWMS, ImageWMS and Tile, which should cover most use-cases for background layers. Nevertheless I'd like to have a config option to set an explicit preview image URL, which "beats" the automatic one.

[...]

PR #223, opened 2021-06-08
Background layer switcher

fschmenger:
This branch adds a background layer switcher control to the map according to #203.

Wegue_bglayerswitcher

A description on how to configure background layers and preview images is available from the updated documentation.


app-starter/static/app-conf-projected.json
```diff
@@ -86,9 +86,8 @@
       "projection": "EPSG:28992",
       "tileGridRef": "dutch_rd",
       "attributions": "<a href='https://www.pdok.nl' target='_blank'>PDOK</a> by Dutch Kadaster",
-      "isBaseLayer": false,
-      "visible": false,
-      "displayInLayerList": true
+      "isBaseLayer": true,
```
chrismayer, line 89: This seems not be an appropriate base layer, this has more an overlay character, so I vote set `isBaselayer: false` here

src/components/bglayerswitcher/LayerPreviewImage.vue
```diff
@@ -0,0 +1,47 @@
...
+      return this.layer.get('previewImage') || LayerPreview.getUrl(
+        this.layer,
+        this.mapView.getCenter(),
+        this.mapView.getResolution(),
+        this.mapView.getProjection()
+      );
+    }
+  }
+};
+</script>
```
chrismayer, line 47: Please add a new line at the end of files

chrismayer (approved):
This is really cool, thanks for this @fschmenger! I like the component a lot. Some questions for clarification:

- Is there a way to hide / disable the bgLayerSwitcher somehow? Or ist it always visible as long as you have one base layer?
- Custom positioning et. al. has to be done by CSS?

Could you please add some basic unit tests? This would be very much appreciated.

chrismayer (approved):
This is really cool, thanks for this @fschmenger! I like the component a lot. Some questions for clarification:

- Is there a way to hide / disable the bgLayerSwitcher somehow? Or ist it always visible as long as you have one base layer?
- Custom positioning et. al. has to be done by CSS?

Could you please add some basic unit tests? This would be very much appreciated.

fschmenger:
Hi Christian, I just returned from a 2 weeks holiday. Thanks for the review, I will look into the requested changes and provide some basic unit tests. Regarding your questions.

* The switcher becomes automatically visible, as soon as 1 base layer is configured.
* Positioning uses CSS (for now). This is obviously not one of the generic modules, as the behavior and placement of the button is custom.

fschmenger:
@chrismayer All requested changes should have been addressed now.

chrismayer:
Thanks for your ongoing effort on this @fschmenger! Regarding your comments:

> The switcher currently is always visible and I haven't introduced any config option to disable it. 

OK for me at least for the moment. Let's have a look how often it is requested not to have the BackgroundSwitchter tool. As a workaround the button could be hidden by CSS in a Wegue application. 

> Positioning uses CSS (for now). This is obviously not one of the generic modules, as the behavior and placement of the button is custom.

Thanks for clarification. This is totally OK and it should be tackled in a bigger view, how to place modules, et. al.

I'll merge now.

PR #224, opened 2021-06-09
Add image based rendering in InfoClickWin

chrismayer:
This adds an (optional) additional image based rendering template in the `InfoClickWin` component. This allows to show an image and a description based on configurable attributes of the clicked feature.
The template is (more or less) the one of the deprecated `FeatureInfoWin`, which never made it as a module. Thus also removes `FeatureInfoWin` component.

The default behavior is as before, so as long as it is not configured actively in the app-conf  the `InfoClickWin` shows the feature properties / attributes and the clicked coordinates in tables.

A valid `"wgu-infoclick"` module configuration to use the image based approach looks like:

```json
    "wgu-infoclick": {
      "target": "menu",
      "win": "sidebar",
      "icon": "info",
      "darkLayout": true,
      "draggable": false,
      "showMedia": true,
      "mediaInfoLinkText": "More Info",
      "mediaInfoLinkUrlProp": "image",
      "imageProp": "image",
      "imageDescriptionProp": "imgTitle"
    },
```  
image

whereas the default config would render the following (as before):

image

A more general approach how to make the rendering for the `FeatureInfo` more flexible should be evaluated separately and in a bigger manner.


JakobMiksch:
@chrismayer I have not tested it but the code looks fine.

chrismayer:
Thanks for the review @JakobMiksch

PR #225, opened 2021-07-01
Map recorder

fschmenger:
This adds a new component to Wegue, which is capable of capturing all visible map layers and creates a video from it.
Video codecs supported by the browser are automatically detected, and the control allows for adjusting video quality related parameters like frame- and bit rate.

wegue_maprecorder

Technical notes:
* The control does not support IE. For browser compatibility see the following [MDN chart](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder#browser_compatibility).
* Due to the [tainted canvas problem](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image#security_and_tainted_canvases) captured layers have to be compliant to the same origin policy. Therefore I added support for the crossOrigin attribute in the layer configuration. All layers with CORS enabled sources have been declared with "crossOrigin"="anonymous" attribute (This is currently all but http://ows.terrestris.de/osm-gray/service and https://ahocevar.com/geoserver/wms - the latter seems to be broken anyway)

### DIFF
--- a/src/components/maprecorder/MapRecorderWin.vue
+++ b/src/components/maprecorder/MapRecorderWin.vue
@@ -181,7 +181,8 @@ export default {
      */
     drawCanvas () {
       var me = this;
-      me.mapContext.clearRect(0, 0, me.mapCanvas.width, me.mapCanvas.height);
+      me.mapContext.fillStyle = 'white';
+      me.mapContext.fillRect(0, 0, me.mapCanvas.width, me.mapCanvas.height);
 
       Array.prototype.forEach.call(
         document.querySelectorAll('.ol-layer canvas'),
--- a/src/components/maprecorder/MapRecorderWin.vue
+++ b/src/components/maprecorder/MapRecorderWin.vue
@@ -81,8 +81,8 @@
           fab
           outlined 
           @click="toggleRecord">
-          <v-icon v-if="recording">mdi-video-off</v-icon> 
-          <v-icon v-else>mdi-video</v-icon> 
+          <v-icon v-if="recording">stop</v-icon> 
+          <v-icon v-else color="red darken-4">fiber_manual_record</v-icon> 
         </v-btn>
         <v-flex fill-height grow> 
           <v-progress-linear
--- a/src/components/maprecorder/MapRecorderWin.vue
+++ b/src/components/maprecorder/MapRecorderWin.vue
@@ -1,0 +1,298 @@
+<template>
+  <wgu-module-card v-bind="$attrs"
+      :moduleName="moduleName"
+      class="wgu-maprecorder-win" 
+      :icon="icon" 
+      :title="title"
+      width=350>
+    
+    <v-expansion-panels :multiple="true" :accordion="true" class="overflow-y-auto">
+      <v-expansion-panel>
+        <v-expansion-panel-header> 
+          <v-layout align-center>
+            <v-icon class="mr-4">settings</v-icon>
+            Optionen
+          </v-layout>
+        </v-expansion-panel-header>
+        <v-expansion-panel-content>
+          <v-card
+            flat
+            color="transparent"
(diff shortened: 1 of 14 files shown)


src/components/maprecorder/MapRecorderWin.vue
```diff
@@ -0,0 +1,299 @@
...
+    title: {type: String, required: false, default: 'Map recorder'}
+  },
+  data () {
+    const mimeTypes = this.getSupportedMimeTypes();
+
+    return {
+      moduleName: 'wgu-maprecorder',
+
+      /**
+      * Custom canvas element for drawing the Open layers map.
```
JakobMiksch, line 130:
```suggestion
      * Custom canvas element for drawing the OpenLayers map.
```

src/components/maprecorder/MapRecorderWin.vue
```diff
@@ -0,0 +1,299 @@
...
+      /**
+       * Context of the custom canvas element.
+       */
+      mapContext: null,
+      /**
+       * Canvas recorder.
+       */
+      recorder: null,
+      /**
+       * Recording state
```
JakobMiksch, line 142:
```suggestion
       * Recording state.
```

JakobMiksch (approved):
Impressive feature!! I tried it and it works well. 
Inline some minor comments regarding formatting.

I like the UI as well. Spontaneously I have the following ideas, which are non-blocking and can be converted into an follow-up issue.

- it would be nice to see the values of the frame rate and the bit rate. 
- Maybe 3 predefined configurations could be provided like "low", "middle" or "high" which already have some useful defaults for non-technical users
- Using a red "record" button and a black square as "stop" button might be more intuitive for users

src/components/maprecorder/MapRecorderWin.vue
fschmenger: Thanks for the review Jakob! I'll hopefully find some time this week to look into the requested changes.

JakobMiksch: @JakobMiksch the points I listed are just ideas that we can convert to future issues
I am totally fine with merging this PR as is

fschmenger: I suggest you wait with merging, since fixing a few formatting issues isn't really worth a separate PR. Regarding your proposals I'll have a look and see what is easy to achieve.

JakobMiksch: I will not merge anyway. let's wait for @chrismayer

fschmenger:
I addressed your proposal regarding icons and code formatting.

Changing the UI to pick a resolution is certainly debatable.
Here were some quick attempts but none spontaneously satisfied me:

* Making the thumb-label persistent overflows in to the label, this is ugly
map_recorder_thumb_label_always

* Using a persistent hint looks OK, but takes a bit too much of screen space for my taste.
map_recorder_hint

I could look into tick labels to provide a sub line with something like {Low, Medium, High}. Another option would be to customize the hint to appear on the right of the slider.
Any suggestions on the UI welcome.

Anyway, as I'm currently in the process of localizing I would kindly ask you to merge it as is, as introducing new texts will result in merge chaos over here. We can then convert it into an issue and I provide the changes later.

JakobMiksch:
@fschmenger thanks for your efforts about the UI. The persistent hint would look okay.
Merging this branch "as is" is totally fine for me. I am very happy that we already have such an elaborate functionality in Wegue. We can do the finetuning with the UI later

chrismayer:
Thanks for this impressive feature @fschmenger! Really cool :+1: 

Thanks for the review @JakobMiksch! I am going to merge now...

fschmenger:
@chrismayer Great, thanks for including the feature!

PR #240, opened 2021-08-27
Add 'wgu-geolocator-layer' to permalink ignore list

chrismayer:
This adds the layer to show the geolocation on the map to the ignore layer list of the permalink module, since it makes no sense to have it in the permalink and the modules are (currently) not restored.

/cc @justb4 @fschmenger

### DIFF
--- a/src/components/ol/PermalinkController.js
+++ b/src/components/ol/PermalinkController.js
@@ -13,7 +13,7 @@ export default class PermalinkController {
   projection = null;
   conf = null;
   urlParams = null;
-  ignoreLayers = ['wgu-measure-layer'];
+  ignoreLayers = ['wgu-measure-layer', 'wgu-geolocator-layer'];
 
   constructor (map, permalinkConf) {
     this.map = map;


fschmenger:
@chrismayer For now we should also add 'wgu-drag-drop-layer' to the ignore list. 

I think a better long term fix for permalink would be to ignore any layer added dynamically on a module level (not via config and factory) - since those most likely wont work.
An implementation would be fairly straightforward:
* The layer factory will add a custom attribute 'usePermalink' to the OL layer list and default this flag to true (maybe with an option turn it off via config).
* Permalink controller evaluates this flag (instead of using static ignore list).

Then application code can decide whether it makes sense to flag its own layers as usable for permalink.

fschmenger:
Be aware, that 'wgu-drap-drop-layer' contains a random suffix, so it's not a one-liner fix. 
I see no problem incorporating the PR. Should we convert the stuff stated above to an issue?

chrismayer:
Yes, I vote for merging this and having an issue for the more general approach. Would you create such an issue @fschmenger ?

fschmenger:
Agreed, merge this for now and I will open an issue.

chrismayer:
Thanks for the issue tracking @fschmenger!

PR #242, opened 2021-08-30
ESLint configuration upgrade

### LINKED ISSUE #152: DeprecationWarning: The 'parserOptions.ecmaFeatures.exp erimentalObjectRestSpread' option is deprecated
`npm run dev` seems to work well, but complains with this message:

```
(node:16126) [ESLINT_LEGACY_OBJECT_REST_SPREAD] DeprecationWarning: The 'parserOptions.ecmaFeatures.exp
erimentalObjectRestSpread' option is deprecated. Use 'parserOptions.ecmaVersion' instead. (found in "st
andard")                                                                                               
(Use `node --trace-deprecation ...` to show where the warning was created)                             
```

fschmenger:
This branch upgrades ESLint configurations and plugins to the newest NPM packages supported by the current ESLint version - ESLlint package itself is not upgraded. This resolves #152.


test/unit/.eslintrc
```diff
@@ -5,5 +5,8 @@
   "globals": {
     "expect": true,
     "sinon": true
+  },
+  "rules": {
+    "no-unused-expressions": "off"
```
chrismayer, line 10: Why needs this rule to be disabled? backwards compatibility to the old version?

chrismayer (approved):
LGTM, thanks @fschmenger. 

One question you might want to answer for clarification before merging.

test/unit/.eslintrc
fschmenger: Hi Christian,
thanks for the review.

You're right, there's a compability issue for the unit tests, in particular chai assertions, e.g.
```
expect(vm.tileGridDefs).to.be.empty;
```
leads to an _non-unused-expression_ warning.
You can read more about it here (https://stackoverflow.com/questions/37558795/nice-way-to-get-rid-of-no-unused-expressions-linter-error-with-chai).

Note that I disabled this rule within the scope of unit tests only, the source code of Wegue itself is not affected.

chrismayer: Thanks for clarification. Makes totally sense. Ready to merge now...

fschmenger:
@chrismayer I just checked and there seem to be some warnings after init:app is run. It looks to me like lint-fix didnt apply the new rules to the app-starter folder. I'm looking into this now and might re-open this PR.

PR #245, opened 2021-09-06
Apply auto-fixes from eslint to app-starter folder.

fschmenger:
This fixes some warnings on built in conjunction with #242. 

Apparently the app-starter folder is currently excluded from linting. Strangely enough after running app:init, eslint will still not complain but the build will throw some warnings. This should be fixed now, sorry for the oversight.


chrismayer (approved):
Thanks @fschmenger.

PR #248, opened 2021-09-06
Background layer switcher autohide

### LINKED ISSUE #227: BackgroundLayerSwitcher: Disable when no Background layers
Currently it it is not possible to disable the BackgroundLayerSwitcher with the config. See [comment](https://github.com/meggsimum/wegue/pull/223#issuecomment-869464003) in the pull request.

I needed to disable it in an application. **Workaround**: remove line with the component reference in `WgueAppTemplate`: https://github.com/meggsimum/wegue/blob/master/app-starter/WguAppTemplate.vue#L37

It would be useful if the switcher is disabled automatically when there is no background layer set

fschmenger:
Automatically hide the background layer switcher, if there are less than 2 base layers configured - resolves #227.

### DIFF
--- a/src/components/bglayerswitcher/BgLayerList.vue
+++ b/src/components/bglayerswitcher/BgLayerList.vue
@@ -70,9 +70,14 @@
       // https://github.com/vuejs/Discussion/issues/394 .The following works in Firefox and Chrome.
       var slideGroup = this.$refs.slideGroup;
       if (slideGroup) {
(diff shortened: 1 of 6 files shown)


JakobMiksch (approved):
Works well and does what it should. Can be merged from my side.
Thanks @fschmenger

chrismayer (approved):
LGTM, thanks.

PR #249, opened 2021-09-06
Map layer options

### LINKED ISSUE #230: Add support for zIndex when customizing layers.
When testing the background layer picker, I realized that we're currently relying on order of layer declaration to define which layers should appear in foreground. This is somewhat inconvenient when 'baseLayers' are declared at the wrong end of the 'mapLayers' list.

* I suggest to generally support a 'zIndex' configuration property for 'mapLayers'.
* There should be some implicit 'zIndex' assigned when the property is missing, such that 'baseLayers' will get a lower zIndex than foreground layers. What numbers should we use here?

fschmenger:
This branch reworks some of the map layer configuration options. There should be no backward compability issues:

* A new option `zIndex` is introduced - resolves #230. Regarding default values, I decided to assign _-1_ to base layers and _0_ to other layers, which seems to be the least invasive approach.
* The `extent` option is now consistently supported across all layer types.
* Filled some gaps in the map layer documentation.

### DIFF
--- a/docs/map-layer-configuration.md
+++ b/docs/map-layer-configuration.md
@@ -10,7 +10,6 @@ The following properties can be applied to all map layer types
 |--------------------|:---------:|---------|
 | **type**           |  Indicator which layer type is configured |  |
 | **lid**            |  Unique identifier for the layer | `"lid": "my-super-wms-layer"`  |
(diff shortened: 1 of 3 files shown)


JakobMiksch (approved):
I tried the Z-Index. It works well. 
The refactoring in the LayerFactory is very useful. 
Ready to merge from my side.
Thanks @fschmenger

PR #250, opened 2021-09-06
View animations

### LINKED ISSUE #238: Generalized support for view animations.
I would like to incorporate my set of view (zoom and pan) animations, which I implemented for the SAGE application.

Here is an example of the "fly-to-animation" which zooms out until the current and the target location are visible and then zooms in on the target (as well know from e.g. Google Maps) :

https://user-images.githubusercontent.com/46027124/128355090-67bc7e21-d1ce-4081-a1a0-8426dc218136.mp4

The idea would be to configure a view animation object in app-conf globally, e.g.
```JSON
viewAnimation: {
   type: 'fly',
   duration: 3000
}
```

and then pool all invocations of 'view.fit()' into a new utility module called 'ViewAnimations'.
Default will always be a simple fit (= no animation), in case the declaration above is missing.
The animations used in SAGE have been inspired by http://openlayers.org/en/latest/examples/animation.html. For a first start I would support { 'fly', 'bounce' and 'none' } as animation types and we can add more later on.

This will be fairly easy to implement, as I already have the module running in another application and will just have to tweak it a little. Currently every animation supports a toLocation and toExtent function and signatures are: 

```
/**
   * Zoom to the given location by using a "xxx" animation.
   * @param {ol.View} view The `ol.View` of the map.
   * @param {ol.Coordinate} location The destination center point to zoom to.
   * @param {Number} duration An optional animation duration.
   * @param {Number} zoom An optional final zoom level.
   * @param {function(complete)} completionCallback An optional notification that the animation has completed.
   */
toLocation: function (view, location, duration,
  zoom, completionCallback) {
}
```

and 

```
/**
   * Zoom to fit the given extent by using a "xxx" animation.
   * @param {ol.View} view The `ol.View` of the map.
   * @param {ol.Extent} extent The destination extent to zoom to.
   * @param {Number} duration An optional animation duration.
   * @param {function(complete)} completionCallback An optional notification that the animation has completed.
   */
toExtent: function (view, extent, duration, completionCallback) {
}
```

fschmenger:
This branch adds generalized view animations to Wegue - implements and resolves #238 proposal.

A new configuration option `viewAnimation` is introduced, to configure animations on application level, e.g.
```JSON
"viewAnimation": {
  "type": "fly",
  "options": {
    "duration": 3000,
    "zoom": 15,
    "maxZoom": 15
  }
}
```

All view zoom / animate / fit invocations are now pooled into a new utility module `ViewAnimations`.
The only exception is permalink, which initially sets the map extent and is left unchanged. 
I worked a little on the signature of `ViewAnimation.to` method, to keep matters simple and support a broad range of uses cases:
* Various types are now accepted as destination.
* Animation options are wrapped by an object, so they can be specific to each animation type. Modules can choose to override the global app-conf options, if they really need to.

```
 /**
   * Zoom to the given destination.
   * @param {ol.View} view The `ol.View` of the map.
   * @param {ol.Coordinate | ol.Extent | ol.geom.Geometry} destination
   *  The destination coordinate, point, geometry or extent to zoom to.
   * @param {*} completionCallback An optional notification that the animation has completed.
   * @param {*} options Optional configuration object for the animation.
   */
  to (view, destination, completionCallback, options) {
  }
```
  
For a first draft, ViewAnimation supports 4 animation types (`'fly'`, `'pan'`, `'bounce'` and `'none'`) and the set can be enhanced easily later on.

Regarding backward compability:
Some of the individual animation options for modules are now deprecated in favor of the `viewAnimation` option. I added migration warnings, if the following are still declared:
* `modules.wgu-geolocator.zoomAnimation`
* ` modules.wgu-geolocator.zoomAnimationDuration`
* `modules.wgu-geolocator.maxZoom`
* `modules.wgu-geocoder.selectZoom`

For more details please refer to the updated documentation and example app-conf templates.

### DIFF
--- a/docs/wegue-configuration.md
+++ b/docs/wegue-configuration.md
@@ -23,6 +23,7 @@ This describes the Wegue application configuration, which is modelled as JSON do
 | **mapLayers**      | Array of map layer configuration objects | See [mapLayers](map-layer-configuration) |
 | projectionDefs     | Array of CRS / projection definition objects compatible to proj4js | See [projectionDefs](wegue-configuration?id=projectiondefs) |
 | tileGridDefs       | Array of tile grid definition objects | See [tileGridDefs](wegue-configuration?id=tilegriddefs) |
+| viewAnimation      | Configuration object for view animations | See [viewAnimation](wegue-configuration?id=viewAnimation) |
 
 ### projectionDefs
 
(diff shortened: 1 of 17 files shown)


docs/wegue-configuration.md
```diff
@@ -127,6 +128,40 @@ Below is an example for such a configuration object:
   }
 ```
 
+### viewAnimation
+
+Map views can be animated, to zoom or pan from the current to a target location, typically after a user action takes place. This can be configured by the property `viewAnimation` in the main Wegue configuration. Per default animations are disabled.
```
JakobMiksch, line 133:
```suggestion
Map views can be animated, to zoom or pan from the current to a target location, typically after an user action takes place. This can be configured by the property `viewAnimation` in the main Wegue configuration. Per default animations are disabled.
```

JakobMiksch (approved):
Impressive feature. Code looks fine and it works for me. 
One minor spelling note inside. 
Otherwise ready to merge from my side.
Thanks @fschmenger

docs/wegue-configuration.md
fschmenger: Hi Jakob. I don't think this should be 'an', because what matters is whether the following word starts with a vowel **phonetically**. See https://www.grammar.com/a-vs-an-when-to-use/

JakobMiksch: Okay. Does not matter :-)

chrismayer (approved):
LGTM, thanks @fschmenger. I am going to merge now.

Issue #253, opened 2021-09-07
Better layout for Attribute Table when located on sidebar


fschmenger:
Currently the UI of Attribute Table is non optimal when `"win": "sidebar"` is specified for the module in app-conf.

![wegue_attribute_table_sidebar](https://user-images.githubusercontent.com/46027124/132346736-0bbb7d7d-6f3b-47ba-8f26-dd73ef40cf18.png)

* The 'choose a layer' v-select box is too wide to fit into the the window header (wgu-win-toolbar) properly. 
* The sidebar already comes with a vertical scollbar, therefore attribute should avoid using one itself.
 
This can be regarded as follow up task to #213.

fschmenger:
Regarding point 1:
It may also be an issue for small screen width devices when not running in sidebar mode. 
@chrismayer @JakobMiksch Would it be a better approach to place the selectunder the header in general?

Regarding point 2:
@chrismayer @JakobMiksch Can we solve this by just using smaller page sizes on the `wgu-attributetable` ?

Issue #243, opened 2021-09-02
Support a windowed option for geocoder


fschmenger:
I propose to introduce support for running the UI of geocoder in a floating and sidebar window. This is a follow up task to standardising the behavior of toolbar-buttons and modules, which was started with #206 and #207.

For a proof of concept implementation I introduced a new module card attribute **"winHeader": false** to disable the window toolbar - name of the attribute is subject to discussion. Here's an example for the module definitions and a corresponding screenshot:

![Wegue_geocoder_floating](https://user-images.githubusercontent.com/46027124/131835159-0708039e-4d5d-40c4-b4aa-0542fc43cd27.png)

```JSON
 "modules": {
    "wgu-geocoder": {
     "target": "toolbar",
      "win": "floating",
      "winHeader": false,
      "icon": "search",
      "initPos": {
        "left": 8,
        "top": 74
      },
      ......
   }
}
```

![Wegue_geocoder_sidebar](https://user-images.githubusercontent.com/46027124/131835166-9df3a78b-404d-4cc5-ae69-b6506d2eae95.png)

```JSON
 "modules": {
    "wgu-geocoder": {
      "target": "toolbar",
      "win": "sidebar",
      "winHeader": false,
      "icon": "search",
      ......
   }
}
```

fschmenger:
@chrismayer 
My question regarding this: Do we want to support placement of the geocoder searchbar on the Wegue toolbar in the future?
 
In my opinion this is not a great solution, because
* it scales bad on devices with small screen width
* all toolbar-buttons currently support the **target": "menu"** option. If this is set for the geocoder, it breaks the layout of the hamburger menu.

Even though I got a prototype to work, a generical switch between placing content on the toolbar and on a card is pretty cumbersome to implement.

PR #262, opened 2021-10-18
Add support for vuetify theming

### LINKED ISSUE #202: Use Vuetify Theming

ntma:
This PR allows overriding the default Vuetify theme configuration.

#### Tasks:
* [x] add app context config to override the default theme;
   * property `baseColor` will be replaced by `colorTheme`; 
* [x] create toolbar tool to switch between light/dark modes;
* [x] write set of default colors (app-config);
  * fallback to default if no colors are provided;
* [x] remove deprecated color configurations (related to `baseColor`);
* [x]  add log message to warn on the deprecated `baseColor`;
* [x] update unit tests;
  * baseColor;
  * colorTheme;
  * darkLayout; 
* [x] update docs;
  * colorTheme property;
  * tips on how to pick theme colors;
* [x] update locales;

#### Extras:
* expose color classes as CSS variables ([docs](https://vuetifyjs.com/en/features/theme/#custom-properties));
* ~~allow~~ Vuetify theme presets (material studies).
  - requires webpack version >= 4

#### To improve in the future
* solo fields ([discussion here](https://github.com/meggsimum/wegue/pull/262#issuecomment-992700598));
* hover over cards ([discussion here](https://github.com/meggsimum/wegue/pull/262#issuecomment-981593703));

#### Material design guidelines
* [Color system](https://material.io/design/color/the-color-system.html);
* [Dark theme](https://material.io/design/color/dark-theme.html);
* [Components](https://material.io/archive/guidelines/material-design/introduction.html#);

Closes #202

### DIFF
--- a/app-starter/WguAppTemplate.vue
+++ b/app-starter/WguAppTemplate.vue
@@ -26,15 +26,15 @@
     <slot name="wgu-before-content" />
     <v-content app>
       <v-container id="ol-map-container" fluid fill-height class="pa-0">
-        <wgu-map color="secondary onsecondary--text" />
+        <wgu-map />
         <!-- layer loading indicator -->
-        <wgu-maploading-status color="secondary" />
+        <wgu-maploading-status />
         <slot name="wgu-after-map"> 
         </slot>
         <!-- Portal to overlay the map content from an application module -->
         <portal-target name="map-overlay" />
         <wgu-app-logo />
-        <wgu-bglayerswitcher color="secondary"/>
+        <wgu-bglayerswitcher />
       </v-container>
     </v-content>
 
--- a/app-starter/WguAppTemplate.vue
+++ b/app-starter/WguAppTemplate.vue
@@ -3,7 +3,7 @@
 
     <slot name="wgu-app-begin" />
 
-    <wgu-app-header :color="baseColor">
+    <wgu-app-header color="primary">
       <!-- forward the slots of AppHeader -->
       <slot name="wgu-tb-start" slot="wgu-tb-start" />
       <slot name="wgu-tb-after-title" slot="wgu-tb-after-title" />
@@ -17,7 +17,7 @@
     <wgu-app-sidebar v-if="sidebarWins.length" v-bind="sidebarConfig">
         <template v-for="(moduleWin, index) in sidebarWins">
           <component
-            :is="moduleWin.type" :key="index" :color="baseColor"
+            :is="moduleWin.type" :key="index" color="primary"
             v-bind="moduleWin"
           />
       </template>
@@ -26,21 +26,21 @@
     <slot name="wgu-before-content" />
     <v-content app>
       <v-container id="ol-map-container" fluid fill-height class="pa-0">
-        <wgu-map :color="baseColor" />
+        <wgu-map color="primary" />
         <!-- layer loading indicator -->
-        <wgu-maploading-status :color="baseColor" />
+        <wgu-maploading-status color="primary" />
         <slot name="wgu-after-map"> 
         </slot>
         <!-- Portal to overlay the map content from an application module -->
         <portal-target name="map-overlay" />
         <wgu-app-logo />
-        <wgu-bglayerswitcher :color="baseColor"/>
+        <wgu-bglayerswitcher color="primary"/>
       </v-container>
     </v-content>
 
     <template v-for="(moduleWin, index) in floatingWins">
       <component
-        :is="moduleWin.type" :key="index" :color="baseColor"
+        :is="moduleWin.type" :key="index" color="primary"
         v-bind="moduleWin"
       />
     </template>
@@ -48,7 +48,7 @@
     <slot name="wgu-before-footer" />
 
     <wgu-app-footer
-      :color="baseColor"
+      color="primary"
       :footerTextLeft="$t('app.footerTextLeft')"
       :footerTextRight="$t('app.footerTextRight')"
       :showCopyrightYear="showCopyrightYear"
@@ -101,8 +101,7 @@
         sidebarConfig: this.getSidebarConfig(),
         floatingWins: this.getModuleWinData('floating'),
         sidebarWins: this.getModuleWinData('sidebar'),
-        showCopyrightYear: Vue.prototype.$appConfig.showCopyrightYear,
-        baseColor: Vue.prototype.$appConfig.baseColor
+        showCopyrightYear: Vue.prototype.$appConfig.showCopyrightYear
       }
     },
     created () {
--- a/app-starter/WguAppTemplate.vue
+++ b/app-starter/WguAppTemplate.vue
@@ -17,7 +17,7 @@
     <wgu-app-sidebar v-if="sidebarWins.length" v-bind="sidebarConfig">
         <template v-for="(moduleWin, index) in sidebarWins">
           <component
-            :is="moduleWin.type" :key="index" color="secondary"
+            :is="moduleWin.type" :key="index"
             v-bind="moduleWin"
           />
       </template>
@@ -40,7 +40,7 @@
 
     <template v-for="(moduleWin, index) in floatingWins">
       <component
-        :is="moduleWin.type" :key="index" color="secondary"
+        :is="moduleWin.type" :key="index"
         v-bind="moduleWin"
       />
     </template>
--- a/app-starter/WguAppTemplate.vue
+++ b/app-starter/WguAppTemplate.vue
@@ -17,7 +17,7 @@
     <wgu-app-sidebar v-if="sidebarWins.length" v-bind="sidebarConfig">
         <template v-for="(moduleWin, index) in sidebarWins">
           <component
-            :is="moduleWin.type" :key="index" color="primary"
+            :is="moduleWin.type" :key="index" color="secondary"
             v-bind="moduleWin"
           />
       </template>
@@ -26,21 +26,21 @@
     <slot name="wgu-before-content" />
     <v-content app>
       <v-container id="ol-map-container" fluid fill-height class="pa-0">
-        <wgu-map color="primary" />
+        <wgu-map color="secondary onsecondary--text" />
         <!-- layer loading indicator -->
-        <wgu-maploading-status color="primary" />
+        <wgu-maploading-status color="secondary" />
         <slot name="wgu-after-map"> 
         </slot>
         <!-- Portal to overlay the map content from an application module -->
         <portal-target name="map-overlay" />
         <wgu-app-logo />
-        <wgu-bglayerswitcher color="primary"/>
+        <wgu-bglayerswitcher color="secondary"/>
       </v-container>
     </v-content>
 
     <template v-for="(moduleWin, index) in floatingWins">
       <component
-        :is="moduleWin.type" :key="index" color="primary"
+        :is="moduleWin.type" :key="index" color="secondary"
         v-bind="moduleWin"
       />
     </template>
--- a/app-starter/WguAppTemplate.vue
+++ b/app-starter/WguAppTemplate.vue
@@ -3,7 +3,7 @@
 
     <slot name="wgu-app-begin" />
 
-    <wgu-app-header color="primary">
+    <wgu-app-header>
       <!-- forward the slots of AppHeader -->
       <slot name="wgu-tb-start" slot="wgu-tb-start" />
       <slot name="wgu-tb-after-title" slot="wgu-tb-after-title" />
@@ -48,7 +48,6 @@
     <slot name="wgu-before-footer" />
 
     <wgu-app-footer
-      color="primary"
       :footerTextLeft="$t('app.footerTextLeft')"
       :footerTextRight="$t('app.footerTextRight')"
       :showCopyrightYear="showCopyrightYear"
--- a/app-starter/static/app-conf-minimal.json
+++ b/app-starter/static/app-conf-minimal.json
@@ -1,6 +1,9 @@
 {
 
-  "baseColor": "green darken-3",
+  "theme": {
+    "dark": false
+  },
+
   "showCopyrightYear": true,
 
   "lang": {
--- a/app-starter/static/app-conf-minimal.json
+++ b/app-starter/static/app-conf-minimal.json
@@ -38,8 +38,7 @@
     "wgu-helpwin": {
       "target": "toolbar",
       "win": "floating",
-      "icon": "help",
-      "darkLayout": true
+      "icon": "help"
     }
   }
 
--- a/app-starter/static/app-conf-minimal.json
+++ b/app-starter/static/app-conf-minimal.json
@@ -1,7 +1,15 @@
 {
 
   "colorTheme": {
-    "dark": false
+    "dark": false,
+    "themes": {
+      "light": {
+        "primary": "#2E7D32"
+      },
+      "dark": {
+        "secondary": "#A5D6A7"
+      }
+    }
   },
 
   "showCopyrightYear": true,
--- a/app-starter/static/app-conf-minimal.json
+++ b/app-starter/static/app-conf-minimal.json
@@ -1,6 +1,6 @@
 {
 
-  "theme": {
+  "colorTheme": {
     "dark": false
   },
 
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -1,7 +1,30 @@
 {
 
   "theme": {
-    "dark": true
+    "dark": false,
+    "themes": {
+      "light": {
+        "primary": "#af2622",
+        "onprimary": "#ffffff",
+        "secondary": "#ec483b",
+        "onsecondary": "#ffffff",
+        "onsurface": "#bf302d",
+        "accent": "#ffffff",
+        "error": "#b00020"
+      },
+      "dark": {
+        "primary": "#272727",
+        "onprimary": "#ffffff",
+        "secondary": "#ea9b9b",
+        "onsecondary": "#272727",
+        "onsurface": "#1e1e1e",
+        "accent": "#ea9b9b",
+        "error": "#cf6679"
+      }
+    },
+    "options": {
+      "customProperties": true
+    }
   },
 
   "logo": "https://dummyimage.com/100x100/aaa/fff&text=Wegue",
@@ -192,14 +215,14 @@
       "target": "menu",
       "win": "floating",
       "icon": "layers",
-      "darkLayout": true,
+      "darkLayout": false,
       "draggable": false
     },
     "wgu-measuretool": {
       "target": "menu",
       "win": "floating",
       "icon": "photo_size_select_small",
-      "darkLayout": true,
+      "darkLayout": false,
       "draggable": false,
       "strokeColor": "#c62828",
       "fillColor": "rgba(198,40,40,0.2)",
@@ -212,7 +235,7 @@
       "target": "menu",
       "win": "floating",
       "icon": "info",
-      "darkLayout": true,
+      "darkLayout": false,
       "draggable": false,
       "initPos": {
         "left": 8,
@@ -266,7 +289,7 @@
       "target": "menu",
       "win": "floating",
       "icon": "table_chart",
-      "darkLayout": true,
+      "darkLayout": false,
       "syncTableMapSelection": true
     },
     "wgu-localeswitcher": {
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -1,6 +1,6 @@
 {
 
-  "theme": {
+  "colorTheme": {
     "dark": false,
     "themes": {
       "light": {
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -192,14 +192,12 @@
       "target": "menu",
       "win": "floating",
       "icon": "layers",
-      "darkLayout": false,
       "draggable": false
     },
     "wgu-measuretool": {
       "target": "menu",
       "win": "floating",
       "icon": "photo_size_select_small",
-      "darkLayout": false,
       "draggable": false,
       "strokeColor": "#c62828",
       "fillColor": "rgba(198,40,40,0.2)",
@@ -212,7 +210,6 @@
       "target": "menu",
       "win": "floating",
       "icon": "info",
-      "darkLayout": false,
       "draggable": false,
       "initPos": {
         "left": 8,
@@ -222,7 +219,6 @@
     },
     "wgu-geocoder": {
       "target": "toolbar",
-      "darkLayout": true,
       "minChars": 2,
       "queryDelay": 200,
       "debug": false,
@@ -234,14 +230,12 @@
       }
     },
     "wgu-zoomtomaxextent": {
-      "target": "toolbar",
-      "darkLayout": true
+      "target": "toolbar"
     },
     "wgu-maprecorder": {
       "target": "toolbar",
       "win": "floating",
       "icon": "mdi-video",
-      "darkLayout": true,
       "draggable": false,
       "initPos": {
         "left": 8,
@@ -251,27 +245,22 @@
     "wgu-helpwin": {
       "target": "toolbar",
       "win": "floating",
-      "icon": "help",
-      "darkLayout": true
+      "icon": "help"
     },
     "wgu-geolocator": {
-      "target": "toolbar",
-      "darkLayout": true
+      "target": "toolbar"
     },
     "wgu-themeswitcher": {
-      "target": "toolbar",
-      "darkLayout": true
+      "target": "toolbar"
     },
     "wgu-attributetable": {
       "target": "menu",
       "win": "floating",
       "icon": "table_chart",
-      "darkLayout": false,
       "syncTableMapSelection": true
     },
     "wgu-localeswitcher": {
-      "target": "toolbar",
-      "darkLayout": true
+      "target": "toolbar"
     }
   }
 }
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -21,9 +21,6 @@
         "accent": "#ea9b9b",
         "error": "#ff6f00"
       }
-    },
-    "options": {
-      "customProperties": true
     }
   },
 
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -258,6 +258,10 @@
       "target": "toolbar",
       "darkLayout": true
     },
+    "wgu-themeswitcher": {
+      "target": "toolbar",
+      "darkLayout": true
+    },
     "wgu-attributetable": {
       "target": "menu",
       "win": "floating",
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -10,7 +10,7 @@
         "onsecondary": "#ffffff",
         "onsurface": "#bf302d",
         "accent": "#ffffff",
-        "error": "#b00020"
+        "error": "#ff6f00"
       },
       "dark": {
         "primary": "#272727",
@@ -19,7 +19,7 @@
         "onsecondary": "#272727",
         "onsurface": "#1e1e1e",
         "accent": "#ea9b9b",
-        "error": "#cf6679"
+        "error": "#ff6f00"
       }
     },
     "options": {
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -1,6 +1,8 @@
 {
 
-  "baseColor": "red darken-3",
+  "theme": {
+    "dark": true
+  },
 
   "logo": "https://dummyimage.com/100x100/aaa/fff&text=Wegue",
   "logoWidth": "100",
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -1,27 +1,7 @@
 {
 
   "colorTheme": {
-    "dark": false,
-    "themes": {
-      "light": {
-        "primary": "#af2622",
-        "onprimary": "#ffffff",
-        "secondary": "#ec483b",
-        "onsecondary": "#ffffff",
-        "onsurface": "#bf302d",
-        "accent": "#ffffff",
-        "error": "#ff6f00"
-      },
-      "dark": {
-        "primary": "#272727",
-        "onprimary": "#ffffff",
-        "secondary": "#ea9b9b",
-        "onsecondary": "#272727",
-        "onsurface": "#1e1e1e",
-        "accent": "#ea9b9b",
-        "error": "#ff6f00"
-      }
-    }
+    "dark": false
   },
 
   "logo": "https://dummyimage.com/100x100/aaa/fff&text=Wegue",
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -251,7 +251,8 @@
       "target": "toolbar"
     },
     "wgu-themeswitcher": {
-      "target": "toolbar"
+      "target": "toolbar",
+      "icon": "dark_mode"
     },
     "wgu-attributetable": {
       "target": "menu",
--- a/docs/wegue-configuration.md
+++ b/docs/wegue-configuration.md
@@ -40,7 +40,6 @@ Example:
       "onprimary": "#ffffff",
       "secondary": "#ec483b",
       "onsecondary": "#ffffff",
-      "accent": "#ffffff",
       "error": "#ff6f00"
     },
     "dark": {    // Dark theme configuration
@@ -48,7 +47,6 @@ Example:
       "onprimary": "#ffffff",
       "secondary": "#ea9b9b",
       "onsecondary": "#272727",
-      "accent": "#ea9b9b",
       "error": "#ff6f00"
     }
   }
@@ -60,16 +58,18 @@ Each theme configuration contains the following classes of colors:
 | Color         | Description         | Example       |
 | ------------- |:-------------------:| ------------- |
 | primary       | color for main UI components             | Header, Footer |
-| secondary     | color to accent selected parts of the UI | Floating buttons, selection controls, progress bars |
-| accent        | accent selected parts of the UI          | Mainly used to accent components in dark mode  |
-| error         | semantic color for errors                | Used in components that display error messages  |
+| secondary     | color to accent selected parts of the UI | Floating buttons, selection controls, progress bars   |
+| info          | semantic color for information           | Used in components that display information messages  |
+| success       | semantic color for success               | Used in components that display success messages      |
+| warning       | semantic color for warning               | Used in components that display warning messages      |
+| error         | semantic color for error                 | Used in components that display error messages        |
 
 
 In addition, Wegue also supports the following "on" colors:
 
 | Color         | Description         | Example       |
 | ------------- |:-------------------:| ------------- |
-| onprimary     | color over primary color   | typography/icons over primary color |
+| onprimary     | color over primary color   | typography/icons over primary color   |
 | onsecondary   | color over secondary color | typography/icons over secondary color |
 
 To simplify the theming configuration, if the "themes" property isn't configured, Wegue will fallback to the default colors in the example above. Otherwise, both the "light" and "dark" themes will be built based on the respective configured colors. The following tables specify which colors are mandatory and their respective default values.
@@ -79,19 +79,23 @@ To simplify the theming configuration, if the "themes" property isn't configured
 | ------------- |:---------:|:-------------------:|
 | primary       | yes |       - |
 | secondary     |  no | primary |
-| accent        |  no | primary |
+| information   |  no | #2196F3 |
+| success       |  no | #4CAF50 |
+| warning       |  no | #FFC107 |
 | error         |  no | #FF5252 |
-| onprimary     |  no | white if primary is a dark color. <br/> black if primary is a light color |
+| onprimary     |  no | white if primary is a dark color. <br/> black if primary is a light color     |
 | onsecondary   |  no | white if secondary is a dark color. <br/> black if secondary is a light color |
 
 #### Dark theme:
 | Color         | Mandatory | Default             |
 | ------------- |:---------:|:-------------------:|
-| primary       |   - |   #272727 |
-| secondary     | yes |         - |
-| accent        |  no | secondary |
-| error         |  no |   #FF5252 |
-| onprimary     |  no | white if primary is a dark color. <br/> black if primary is a light color |
+| primary       |   - | #272727 |
+| secondary     | yes |       - |
+| information   |  no | #2196F3 |
+| success       |  no | #4CAF50 |
+| warning       |  no | #FFC107 |
+| error         |  no | #FF5252 |
+| onprimary     |  no | white if primary is a dark color. <br/> black if primary is a light color     |
 | onsecondary   |  no | white if secondary is a dark color. <br/> black if secondary is a light color |
 
 Note that there is a clear asymmetry in the "light" and "dark" theme configuration. In the "light" theme, the primary color is mandatory and all the 
--- a/docs/wegue-configuration.md
+++ b/docs/wegue-configuration.md
@@ -40,7 +40,6 @@ Example:
       "onprimary": "#ffffff",
       "secondary": "#ec483b",
       "onsecondary": "#ffffff",
-      "onsurface": "#bf302d",
       "accent": "#ffffff",
       "error": "#ff6f00"
     },
@@ -49,7 +48,6 @@ Example:
       "onprimary": "#ffffff",
       "secondary": "#ea9b9b",
       "onsecondary": "#272727",
-      "onsurface": "#1e1e1e",
       "accent": "#ea9b9b",
       "error": "#ff6f00"
     }
@@ -73,7 +71,6 @@ In addition, Wegue also supports the following "on" colors:
 | ------------- |:---------:|:-------------------:| ------------- |
 | onprimary     | yes | color over primary color   | typography/icons over primary color |
 | onsecondary   | yes | color over secondary color | typography/icons over secondary color |
-| onsurface     | yes | color over surface color   | background for selection controls over primary color  |
 
 
 ### projectionDefs
--- a/docs/wegue-configuration.md
+++ b/docs/wegue-configuration.md
@@ -94,6 +94,8 @@ To simplify the theming configuration, if the "themes" property isn't configured
 | onprimary     |  no | white if primary is a dark color. <br/> black if primary is a light color |
 | onsecondary   |  no | white if secondary is a dark color. <br/> black if secondary is a light color |
 
+Note that there is a clear asymmetry in the "light" and "dark" theme configuration. In the "light" theme, the primary color is mandatory and all the 
+others are derived from that. In the "dark" theme, the predominant color must be a shade of black (to comply with the [material design specification](https://material.io/design/color/dark-theme.html)), so the primary color is locked to `#272727`. Therefore, in this case the colors are derived from the second predominant color.
 
 ### projectionDefs
 
--- a/docs/wegue-configuration.md
+++ b/docs/wegue-configuration.md
@@ -75,6 +75,7 @@ In addition, Wegue also supports the following "on" colors:
 To simplify the theming configuration, if the "themes" property isn't configured, Wegue will fallback to the default colors in the example above. Otherwise, both the "light" and "dark" themes will be built based on the respective configured colors. The following tables specify which colors are mandatory and their respective default values.
 
 #### Light theme:
+
 | Color         | Mandatory | Default             |
 | ------------- |:---------:|:-------------------:|
 | primary       | yes |       - |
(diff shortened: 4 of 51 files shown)


src/locales/de.json
```diff
@@ -68,6 +68,8 @@
     "frameRate": "Bildrate (Bilder/s)",
     "bitRate": "Bitrate (MBits/s)",
     "fileName": "Dateiname",
+    "start": "Starten",
+    "stop": "Halt",
```
JakobMiksch, line 72:
```suggestion
    "start": "Start",
    "stop": "Stop",
```

src/components/themeswitcher/ThemeSwitcher.vue
```diff
@@ -0,0 +1,47 @@
+<template>
```
JakobMiksch, line 1: Could you add a short description what this component does?

docs/wegue-configuration.md
```diff
@@ -24,6 +24,56 @@ This describes the Wegue application configuration, which is modelled as JSON do
 | projectionDefs     | Array of CRS / projection definition objects compatible to proj4js | See [projectionDefs](wegue-configuration?id=projectiondefs) |
 | tileGridDefs       | Array of tile grid definition objects | See [tileGridDefs](wegue-configuration?id=tilegriddefs) |
 
+### theme
+
+Wegue supports the Vuetify (light and dark) theme configuration out of the box. This property expects a json object with the same format described in the [vuetify documentation](https://vuetifyjs.com/en/features/theme/#customizing).
+
+Example:
+```json
```
JakobMiksch, line 32: are all 7 colors per theme compulsory?

app-starter/WguAppTemplate.vue
```diff
@@ -3,7 +3,7 @@
```
JakobMiksch, line 3: a cosmetic remark:
the map controls like basemap switcher or zoom buttons should have the primary color as well - as least we had this for our previous projects. See Screenshots:

Basemap Switcher:
image

Zoom Buttons:
image

app-starter/static/app-conf.json
```diff
@@ -1,6 +1,28 @@
 {
 
-  "baseColor": "red darken-3",
+  "theme": {
```
JakobMiksch, line 3: could we rename it to `colorTheme`? 
This would be easier to understand IMHO

app-starter/static/app-conf-minimal.json
```diff
@@ -1,6 +1,9 @@
 {
 
-  "baseColor": "green darken-3",
+  "theme": {
```
JakobMiksch, line 3: Is it possible to not provide the `theme` property at all?
As fallback a default color scheme could be chosen ...

JakobMiksch (commented):
Thanks @ntma 
I tried it and it works well. The appearance of the application really benefits from this new scheming approach. See some (mostly cosmetic) remarks inline.

src/locales/de.json
JakobMiksch: In this case the translation is easy :sweat_smile:

JakobMiksch:
> Hi @JakobMiksch,
> 
> I added a few more fixes to account color schemes that use a light color as primary color. The default color for text/icons is white but in this case we want it to be a dark tone for the primary components.
> 
> Also added a section to the documentation to explain how to use the "theme" config in the `app.config.json`. Should I also write a brief tutorial on how to pick colors for the colors classes that Wegue has?
> 
> ps: whenever you have some free time, would you please review these [two translations](https://github.com/meggsimum/wegue/pull/262/commits/c6357e59f14227ca21e265d7b7e8bacc701a0884)? The context of "start" and "stop" is start recording / stop recording.

It would be great to have a small explanation in the docs how to create color schemes. Or where generate them.

src/components/themeswitcher/ThemeSwitcher.vue
ntma: Yes sure. In the documentation or as a comment in the template?

docs/wegue-configuration.md
ntma: To fully colorize the app we need the `primary`, `secondary` and `accent`. The remaining are occasionally needed in case we set a light background with white text.

I'll include that detail in the documentation too.

app-starter/WguAppTemplate.vue
ntma: Material design specifies that any floating button should take an accent color, so that they are distinguished from the other UI elements (in this case we use the secondary color).

In the end, both the primary and secondary can be set to the same colors, which is the effect that you are describing.

app-starter/static/app-conf-minimal.json
ntma: Yes, I was thinking about that too. Wegue will fallback to it's red theme if no theme is provided :+1:

src/components/themeswitcher/ThemeSwitcher.vue
JakobMiksch: just as a short comment in the template

app-starter/WguAppTemplate.vue
JakobMiksch: alright :+1:

src/components/themeswitcher/ThemeSwitcher.vue
ntma: Done!

app-starter/WguAppTemplate.vue
ntma: In the latest commit, the default primary and secondary colors are the same tone of red :+1:

app-starter/static/app-conf-minimal.json
ntma: In the latest commit, Wegue will fallback to it's default theme configuration.

If there is a `colorTheme` in the config, it will be merged with the default (shallow merge). This means that users will be able to override the property `dark` and `themes`, but not the inner properties of `themes`.

fschmenger:
Hi @ntma.

First of all this is some awesome work and I'm happy to see you got going!
I finally found some time to look over the code briefly and do a little testing. Here are my findings so far:

1.  I think we should streamline the theming / color related attributes in the component implementations. (@chrismayer This is also a question to the PM) Since we now have multiple colors to manage, 1 configurable color for components is no longer sufficient. My proposal would be to hardwire all the mappings between colors and the theme in the component implementations itself and remove color related component attributes entirely. This will also clean up the app template, which currently is cluttered with piping through color definitions.
* This also applies to some specific attributes, e.g.'selColor' in BgLayerList, which are still present but no longer in use. Maybe you could re-check, so there are no unused leftovers.
2. I'm a little bit confused: Should the toolbarbuttons on the app-bar now appear in the secondary color? I currently don't see this happening. But maybe this is also not desired?!
3. I realized you changed some of the button shapes, e.g. BgLayerSwitcher. The idea behind using toggle groups was to work towards all buttons having the same look and feel (e.g. pressed shape). 
4. What is the status quo about removing the 'darkLayout' module configuration property (corresponds to the 'dark' attribute in the components)? Can this now be handled generically by your new theme switcher? In this context I realized that the panel of the hamburger no longer uses the primary color. Is this something the material design specification asks us to do? The idea behind changing it to the color of the app-bar was, that no dark layout wise destinction has to be made, whether a button is located on the app-bar or hamburger menu. 
5. Maybe you could introduce a backward compability warning that the 'baseColor' configuration property is no longer supported (see migrateAppConfig in main.js)
6. For merging the configured theme with the default theme you could also use util/Object.js/mergeDeep - however I'm not sure if this make sense here. Maybe it's a better approach to default 'secondary' and 'accent' colors to the 'primary' if the latter are not configured?!
7. I propose you could add your new settings, e.g. theme switcher to the other application contexts (app-conf-projected and app-conf-projected - but not to app-conf-minimal)

Thats it from me, sorry for the lengthy post.

ntma:
Hi @fschmenger ,

I'll try to make it short :+1: 

> 1. I think we should streamline the theming / color related attributes in the component implementations. (@chrismayer This is also a question to the PM) Since we now have multiple colors to manage, 1 configurable color for components is no longer sufficient. My proposal would be to hardwire all the mappings between colors and the theme in the component implementations itself and remove color related component attributes entirely. This will also clean up the app template, which currently is cluttered with piping through color definitions.

Yes we can do this. I wasn't sure at which level Wegue should control it's colors so I tried to set the control to the top level: `appHeader`, `appFooter` and the map.

> 2.` I'm a little bit confused: Should the toolbarbuttons on the app-bar now appear in the secondary color? I currently don't see this happening. But maybe this is also not desired?!

I am not sure if you are referring to the panel of the hamburger menu in the app-bar. If so, the answer to 4. may also answer this question.

> 3. I realized you changed some of the button shapes, e.g. LocaleSwitcher. The idea behind using toggle groups was to work towards all buttons having the same look and feel (e.g. pressed shape).

Hmm, but the LocaleSwitcher functions as a drop-down menu while the others (e.g: Recorder) activate a feature. That's the logic that I was following. I was just trying to give an extra and of course we can revert this change at any time :+1: 

> 4. What is the status quo about removing the 'darkLayout' module configuration property (corresponds to the 'dark' attribute in the components)? Can this now be handled generically by your new theme switcher? In this context I realized that the panel of the hamburger no longer uses the primary color. Is this something the material design specification asks us to do? The idea behind changing it to the color of the app-bar was, that no dark layout wise destinction has to be made, whether a button is located on the app-bar or hamburger menu.

The `darkLayout` is now controlled generically by Vuetify. In my perspepctive, the `baseColor` and `darkLayout` acted like twins in the component configuration. If we streamline one, the other should go with it too. If this isn't desired, we can revert the change.

Regarding the hamburger menu and knowing that:
* primary color should be used in the main components (header, footer, brand logo);
* secondary color should be used to accent controls;
* surface color should be used as the background of panels;
then the panel should have a surface color as background with a secondary color for the buttons.

There is a short insight about this topic in the Primary Colors, Secondary Colors and Surface Colors sections of [this material design doc](https://material.io/design/color/the-color-system.html#color-theme-creation).

> 5. Maybe you could introduce a backward compability warning that the 'baseColor' configuration property is no longer supported (see migrateAppConfig in main.js)

Sure thing, I will add this.

> 6. For merging the configured theme with the default theme you could also use util/Object.js/mergeDeep - however I'm not sure if this make sense here. Maybe it's a better approach to default 'secondary' and 'accent' colors to the 'primary' if the latter are not configured?!

Interesting, I didn't know about the `mergeDeep`. Either way, seems diplomatic solution :+1:  My concern is that a user only sets colors for the light theme. In which case the dark theme will fallback to the default blue of Vuetify...

> 7. I propose you could add your new settings, e.g. theme switcher to the other application contexts (app-conf-projected and app-conf-projected - but not to app-conf-minimal)

Sure, I will do this.

chrismayer:
Hi @ntma, big thanks for your ongoing work on this. Looks very very promising :+1: 

I agree that if we introduce theming by vuetify everything should be handled by it. Concurrent approach by having some props set in the app-config outside of the theming config block would cause confusion.

fschmenger:
Hi @ntma, thanks for the clarifications and links to the notable sources.

Since we have a response from the PM, I think we're good to finalize the topics mentioned in 1) and 4) and control colors at a theme level only. As a summary
* Please remove the `darkLayout` property from app-configs. For the color wise distinction between a button being located on the app-bar or hamburger menu (I will ask the PM for cofirmation further below), you could possible do something like this in `AppHeader.vue::getModuleButtons`
```
     getModuleButtons (target) {
      const appConfig = Vue.prototype.$appConfig || {};
      const modulesConfs = appConfig.modules || {};
      let buttons = [];
      for (const key of Object.keys(modulesConfs)) {
        const moduleOpts = appConfig.modules[key];
        if (moduleOpts.target === target) {
          buttons.push({
            type: moduleOpts.win ? 'wgu-toggle-btn' : key + '-btn',
            moduleName: key,
            // Relay the parent control to the button for coloring.
            // Or alternatively compute a 'color' property here based on target==='toolbar',
            // this will avoid redundancies in the component implementations until we have an abstraction for all toolbarbuttons - see #260
            target: target, 
            ...moduleOpts
          });
        }
      }
      return buttons;
    }
```
* Otherwise remove `color` (this also involves similar color related attributes and default color values ) and `dark` props from the components. 
* Remove all other references/initializations to the properties described above from the app-templates.
* Update documentation and unit tests accordingly.

Regarding 2) I misread the code ;). I can see you are using the primary color for all the buttons on the app-bar now, which is what I expected.

Otherwise I can see you're pretty much there!

@chrismayer The remaining questions to the PM:
Regarding 4) Are we good to change the background color of the hamburger menu back to 'surface color' ? This is probably a bit opionated, as the hamburger menu is a special dropdown which can also be seen as an extension to the app-bar (and also requires some coloring switches in the code). However I can see the point of Nuno's recommendation.
Regarding 3) You should make a clear ruling when buttons should have a toggle/pressed state. See Nuno's comment above.

ntma:
Hello all,

@chrismayer Glad that the PR is meeting the expectations! :+1: I will try my best to finish it as soon as possible, since I will be a bit busy for the next days.

@fschmenger

> Since we have a response from the PM, I think we're good to finalize the topics mentioned in 1) and 4) and control colors at a theme level only.

I will advance to this change then :+1: 

> Regarding 2) I misread the code ;). I can see you are using the primary color for all the buttons on the app-bar now, which is what I expected.

No problem! I was all over the code for this PR so some changes may not be that clear.

ntma:
@fschmenger Indeed, thanks for noticing that!

I had it in my notes but eventually forgot to commit that explanation.

docs/wegue-configuration.md
ntma: Hi @JakobMiksch,

In the recent changes, only the "primary" color is mandatory for the "light" theme and the "secondary" for the "dark" theme. The reason between this asymmetry is explained in the [docs](https://github.com/meggsimum/wegue/pull/262/files#diff-c012d313a14a5074d6d5d8d996f6c0e4188266068540fab41390c2200d1282ecR95).

Feel free to reopen this review if needed :+1:

ntma:
Hi guys,

Finally I believe that this PR is coming to it's conclusion. In the last changes:
* the 'onsurface' color was removed . This class was being applied to components that were clearly not over the 'surface';
* the 'accent' color was moved to be handled internally by Wegue. This is a color class used to highlight focused fields (v-text-field/v-select) in the dark mode. The reasons for this change is: the occasional use of the class; and the fact that this class will in most cases be the same color as the secondary (which is already an accent color).
* removed occurrences of the deprecated `baseColor` and `darkLayout`. From now on, components handle the color/dark internally. Except for the components that can be placed in the toolbar or hamburger menu (since the toolbar uses a primary color and the hamburger a secondary).
* added the remaining semantic colors (info/success/warning). Wegue isn't using any of these but I believe that sooner or later they will be needed.
* added documentation on picking colors for the "light"/"dark" themes;
* updated unit tests;

Any question please let me know. Critics are also welcome :+1:

fschmenger:
Hi @ntma,
I gave it an extended code review and test today:
I can see it's coming to a conclusion and I really like most of the recent changes, e.g. I can see clear pattern now, at which level color values are assigned. The documentation is also nice and verbose. 

Most of the remaining findings are minor (lots of text though ;)
1. I ran a little grabbing for css colorization and hardcoded colors. It would be nice to tweak these rules in the process of theming. Maybe you could have a look, since you already have expert knowledge there.
* I noticed hovering on the sheet of the background-layer currently does nothing for dark mode. Maybe theres a better generalistic solution for the following, so the color values can be removed?
https://github.com/meggsimum/wegue/blob/5c4e8d8020e85ba0e6bc39f5b00361f7f4ce30ca/src/assets/css/wegue.css#L35-L39 
* Can we use the standard surface color for the tooltip, e.g. something along the lines of assigning `v-sheet` as class ? https://github.com/meggsimum/wegue/blob/5c4e8d8020e85ba0e6bc39f5b00361f7f4ce30ca/src/components/ol/Map.vue#L442-L443
* Remove uncommented block in https://github.com/meggsimum/wegue/blob/5c4e8d8020e85ba0e6bc39f5b00361f7f4ce30ca/src/components/geocoder/Geocoder.vue#L213-L222
* This rule wont work without `!important` in a production build (footer link color broken)
https://github.com/meggsimum/wegue/blob/5c4e8d8020e85ba0e6bc39f5b00361f7f4ce30ca/app-starter/components/AppFooter.vue#L35-L38
* Probably the help window link should use a default color, as this is not very theme compliant. I tried removing `red--text darken3` but this won't work in production build again. Maybe links are worth a review in general.
https://github.com/meggsimum/wegue/blob/5c4e8d8020e85ba0e6bc39f5b00361f7f4ce30ca/src/components/helpwin/HelpWin.vue#L19
2. I noticed that hover colors of the toolbarbuttons on the app-bar are now either gone or hardly visible, at least when using the default style. Anyway in disregard of the theme, toggle-buttons now have another hover color than the other buttons, which I find a bit disturbing (check out the behavior before the theme change).
3. Personally I would prefer to factor out the following code (including fallbacks) into `ColorThemeUtil`, so we keep the codesize of `main.js` to a minimum.
https://github.com/meggsimum/wegue/blob/5c4e8d8020e85ba0e6bc39f5b00361f7f4ce30ca/src/main.js#L51-L89
4. I'm not sure if I like the following code block, as this will be ackward to replicate for other components using a `v-combobox `(similarly for `AttributeTableWin`). But then it might be to the specific nature of where the controls are placed.. Maybe you could re-check if there is a more straightforward solution? If not it's probably a good idea to wrap up commonly used computed properties for theming (mainly `isDark`, `forceDark` and the `hsla()` values) in a mixin. See src/mixins
https://github.com/meggsimum/wegue/blob/5c4e8d8020e85ba0e6bc39f5b00361f7f4ce30ca/src/components/geocoder/Geocoder.vue#L8-L13
5. Regarding the color of hamburger menu panel (already mentioned this in my last review): If i got the [material design spec](https://material.io/design/color/applying-color-to-ui.html#sheets-and-surfaces) correctly, then we're free to do there whatever we like. Not that I like one color better than the other, but this would make the code a lot simpler to maintain, as we could remove `color` attributes and switches from toolbarbuttons entirely. Waiting for the PM @chrismayer to have a word on this.
> Modal sheets
> Use contrasting colors on surfaces that appear on-screen temporarily, such as navigation drawers and bottom sheets. Usually these surfaces are white, but you can **use your app’s primary or secondary color**.
 


Finally a big up to you. Overall a great job and big step forward for Wegue!

chrismayer:
Hi @ntma, very big thanks to you for ongoing big work! This use huge and an important step for Wegue. Since things are really busy here I "only" checked out your branch and checked the visual results in the apps. Looks really, really good! Here some remarks from my side:

**TODO** 

  - dark mode switcher needs to be a ToggleButton IMHO, so it gets clearer for the user that dark mode can be switched on/off 
  - The colors of the other apps (`appCtx=projected` (orange primary :netherlands: ) and `appCtx=minimal` (greenish primary) ) need to be restored
  
**To be discussed**
  - The headers of the module windows are hard to recognize in the default dark mode (since body color is quite similar), is this intentionally?
  - Colors of the buttons in the vertical toolbar (hamburger menu) differ from the ones in the toolbar. Intentionally? I think this is what @fschmenger mentioned in his [comment](https://github.com/meggsimum/wegue/pull/262#issuecomment-973094119). 

Caution: I am neither designer nor a MD expert. If the two point above have to be like they are in MD, please ignore them. 
  
**Important**
  - All color configurations should now be done by theming with cool defaults (as stated in my last comment in here). From my quick check this seems to be done at most points. So for all of us, if someone spots a UI color not declared by themes, please report.

Regarding the code: Since I doubt that I find a slot for an extensive code review in the next weeks and I don't want to block this: I am fine if @fschmenger and @JakobMiksch (and of course anyone else who is willing to do a review) give their OK.

JakobMiksch (commented):
Looks great. It feels like a  completely new application. Inline some (mostly minor) remarks.

ntma:
Hi guys,

First of all, big thanks for the effort put into reviewing the changes of this PR :+1: 

The majority of the requested changes are clear so I'll proceed on applying them.

I just have a couple of replies for @fschmenger and @chrismayer.

### For @fschmenger 

> Probably the help window link should use a default color, as this is not very theme compliant. I tried removing red--text darken3 but this won't work in production build again. Maybe links are worth a review in general.

Grr, that link escaped the refactoring. When you suggest using a default color, are you referring to any of the color classes that this PR adds (e.g: primary/secondary)?

### For @chrismayer

> The headers of the module windows are hard to recognize in the default dark mode (since body color is quite similar), is this intentionally?

From the content that I researched, I believe it's intentional. There is a brief explanation about this in [this material design documentation](https://material.io/design/color/dark-theme.html#anatomy). Of course this serves only as guidelines so we are free to adjust the shades of grey to provide a clear contrast between headers and body.

> Colors of the buttons in the vertical toolbar (hamburger menu) differ from the ones in the toolbar. Intentionally? I think this is what @fschmenger mentioned in his comment.

Yes it's intentional. The material design documentation suggests the use of the "secondary" color to accent controls that can be interacted with. But again, this is a suggestion, so I believe that we are free to make some exceptions.

src/util/ColorTheme.js
```diff
@@ -0,0 +1,125 @@
...
+const LIGHT_WARNING = '#FFC107';
+const LIGHT_BLACK = '#000000';
+const DARK_WHITE = '#ffffff';
+const DARK_ERROR = '#FF5252';
+const DARK_INFO = '#2196F3';
+const DARK_SUCCESS = '#4CAF50';
+const DARK_WARNING = '#FFC107';
+const DARK_BLACK = '#272727';
+
+/**
```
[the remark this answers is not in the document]
ntma: Hi @JakobMiksch, what kind of formatting is expected here?

chrismayer:
@ntma Thanks for your feedback. Regading your questions towards me:

>>The headers of the module windows are hard to recognize in the default dark mode (since body color is quite similar), is this intentionally?

>From the content that I researched, I believe it's intentional. There is a brief explanation about this in this material design documentation. Of course this serves only as guidelines so we are free to adjust the shades of grey to provide a clear contrast between headers and body.

OK, then it is fine for me.

>>Colors of the buttons in the vertical toolbar (hamburger menu) differ from the ones in the toolbar. Intentionally? I think this is what @fschmenger mentioned in his comment.

>Yes it's intentional. The material design documentation suggests the use of the "secondary" color to accent controls that can be interacted with. But again, this is a suggestion, so I believe that we are free to make some exceptions.

No, that is fine for me. Was just curious how this all works together.

So no further work regarding my 2 questions IMHO.

fschmenger:
Hi @ntma

> Grr, that link escaped the refactoring. When you suggest using a default color, are you referring to any of the color classes that this PR adds (e.g: primary/secondary)?

Yes, I think link colors (as all colors in Wegue) should now be handled by theming. I'm not sure what the material design spec has to say on colorizing links. Maybe use a secondary color or just the text-color from the parent block without any specific changes? Hardcoding colors to red will apparently have very undesirable effects depending on the used theme.
Whatever solution you come up with, I recommend testing your final solution in a production build, as something goes wrong there currently with links.

ntma:
Hi @fschmenger ,

> Yes, I think link colors (as all colors in Wegue) should now be handled by theming. I'm not sure what the material design spec has to say on colorizing links. Maybe use a secondary color or just the text-color from the parent block without any specific changes? Hardcoding colors to red will apparently have very undesirable effects depending on the used theme.
Whatever solution you come up with, I recommend testing your final solution in a production build, as something goes wrong there currently with links.

Yes I agree with that. I will search the MD documentation for any reference on the links colors... but my guess is the same as yours. Perhaps we can use the `secondary--text`, which is a class that is automatically generated by Vuetify using the secondary color. 

Regarding the `all colors in Wegue should now be handled by theming`, I am afraid there will be some that we won't be able to control unless we override some of the Vuetify CSS classes (at least with Vuetify 2.x). I was searching a way of controlling the "surface" color via Vuetify, and I bumped into [this issue](https://github.com/vuetifyjs/vuetify/issues/9873).

fschmenger:
Hi @ntma,

> Regarding the `all colors in Wegue should now be handled by theming`, I am afraid there will be some that we won't be able to control unless we override some of the Vuetify CSS classes (at least in Vuetify 2.x).

Just clearify my point. I'm not talking about exposing surface color configuration etc. to theming, as this is currently not supported - if I got you right !? What I was intending to do, is getting rid of custom css colorization code, by e.g. mapping controls like custom tooltips to standard vuetify controls, so the implicitly defined vuetify surface colors will be used for those. In an ideal world we would have no more css code to customize colors. If this is not possible for one or the other reason, you could probably mark the codeblocks pointed out by me with some `TODO Review ccs colorization`

ntma:
@chrismayer The last commit restored the colors for `app-conf-projected` (dark orange) and `app-conf-minimal` (dark green) (result for both light and dark variants in the sections bellow).

Only the primary color was restored (the secondary color will take the same value as the primary). Let me know if we need to adjust the color tones or add secondary colors to the configuration :+1: 

#### Projected

Screenshot 2021-11-25 at 11-55-40 Wegue Demo App

Screenshot 2021-11-25 at 11-56-19 Wegue Demo App

#### Minimal

Screenshot 2021-11-25 at 11-57-37 Wegue Demo App

Screenshot 2021-11-25 at 11-57-24 Wegue Demo App

ntma:
@fschmenger 

> Just clearify my point. I'm not talking about exposing surface color configuration etc. to theming, as this is currently not supported - if I got you right !? What I was intending to do, is getting rid of custom css colorization code, by e.g. mapping controls like custom tooltips to standard vuetify controls, so the implicitly defined vuetify surface colors will be used for those. In an ideal world we would have no more css code to customize colors. If this is not possible for one or the other reason, you could probably mark the codeblocks pointed out by me with some TODO Review ccs colorization

Yes, you got me right. Sure, as a last resort I will mark the blocks :+1:

fschmenger:
@ntma 
Just looking at the restored color scheme (b01e36265ffd2cef379be8105fcdba9106e8fedd). `Minimal` looks good to me, for `projected` we had `white` for buttons and primary text.

wegue_projected_color_scheme

ntma:
@fschmenger 
> Just looking at the restored color scheme (b01e362). Minimal looks good to me, for projected we had white for buttons and primary text.

In this case it seems the orange is not dark enough for the text to be white. This is controlled by our [brightness function](https://github.com/meggsimum/wegue/pull/262/files#diff-aebae856dd2465df84734b6af75842391520fd963e435e40218a29dbbb1e9c23R21) when there is not onprimary/onsecondary colors defined in the colorTheme.

I see three options here:
* we change the orange to a darker tone;
* we set both "onprimary" and "onsecondary" to white in the app-conf-projected.json;
* we adjust the `checkLuminance` code, although this one will be hard to justify;

fschmenger:
@ntma 
I think we should go with the simplest approach:
> * we set both "onprimary" and "onsecondary" to white in the app-conf-projected.json;

If there are downsides from doing that, I suggest you have a word with @chrismayer.

fschmenger:
I just re-checked the following code-block from my previous comment (1)

https://github.com/meggsimum/wegue/blob/5c4e8d8020e85ba0e6bc39f5b00361f7f4ce30ca/src/assets/css/wegue.css#L35-L39

I didn't remember at that time of writing this but I introduced this rule to give the button on the map an OL like look and feel (behaviour similiar to the zoom buttons). Since this is a subject of further CRs - see sketch from @chrismayer in #266 - we're good to leave that as is for now.

ntma:
@fschmenger 
I was trying to fix the hovering on the BgLayerList and ended up with a solution that uses `v-hover` with a `v-overlay`. ([code here](https://github.com/meggsimum/wegue/pull/262/commits/b1e76848ecda65edd0a45ab9b9419ebd61bd95d3#diff-f1cfe4b715fab02b27e6df328dfb20e491a61c103382e88dbb27c8add7f19979L16))

It is hard to justify the 0.2 opacity with the color="grey", but it seems to apply a similar overlay that is applied with hovering vuetify buttons. Is this acceptable for Wegue?

ntma:
Hi @fschmenger ,

> I'm not sure if I like the following code block, as this will be ackward to replicate for other components using a v-combobox (similarly for AttributeTableWin). But then it might be to the specific nature of where the controls are placed.. Maybe you could re-check if there is a more straightforward solution? If not it's probably a good idea to wrap up commonly used computed properties for theming (mainly isDark, forceDark and the hsla() values) in a mixin. See src/mixins

Regarding the theming for the Geocoder and the AttributeTableWin I ended up using the `solo`, `filled` and `dark` properties of the "v-select" and "v-combobox" to create the effect of a solo field (more info about "Solo Fields" on toolbars in [this material doc](https://material.io/archive/guidelines/components/text-fields.html#text-fields-field-variations)).

In theory, the `solo` property should be enough to create the desired effect for both light and dark variants... but for some reason, when we add solo in the light theme variant, the field background is forced into a pure white color.

Screenshot 2021-11-26 at 11-02-25 Wegue Demo App

Hence the reason why we use the `filled` for the light variant and `solo` for the dark.

Screenshot 2021-11-26 at 11-03-29 Wegue Demo App

Screenshot 2021-11-26 at 11-03-50 Wegue Demo App

The [last commit](https://github.com/meggsimum/wegue/pull/262/commits/5af57905ea5c882aa27ea6167c11f9ab7c17f971) also adds a mixin to expose utilities related to theming. Such as:
* isDarkTheme - checks if the theme is in dark mode;
* isPrimaryDark - checks if the primary color is a dark color;

src/components/geocoder/Geocoder.vue
```diff
@@ -1,34 +1,42 @@
...
+    outlined
+    dense
+    color="accent"
+    :dark="forceDark"
+    :style='{ 
+      display: (hideSearch ? "none" : "block" ),
+      background: isDark
+        ? "hsla(0, 0%, 0%, 0.16)"
+        : "hsla(0, 0%, 100%, 0.04)"
+    }'
```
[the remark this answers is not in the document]
ntma: Hi @JakobMiksch, I just changed this code to remove the conditional `hsla`. See [this](https://github.com/meggsimum/wegue/pull/262#issuecomment-979890616) comment for more info.

src/components/attributeTable/AttributeTableWin.vue
```diff
@@ -8,9 +8,22 @@
...
       <v-select
         v-model="selLayer"
+        color="accent"
+        :dark="forceDark"
+        item-color="secondary"
+        :style='{
+            background: isDark
+              ? "hsla(0, 0%, 0%, 0.16)"
+              : "hsla(0, 0%, 100%, 0.04)"
+        }'
```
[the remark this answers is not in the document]
ntma: Hi @JakobMiksch, I just changed this code to remove the conditional `hsla`. See [this](https://github.com/meggsimum/wegue/pull/262#issuecomment-979890616) comment for more info.

src/locales/de.json
```diff
@@ -88,5 +90,9 @@
 
   "wgu-localeswitcher": {
     "title": "Spracheinstellungen"
+  },
+
+  "wgu-themeswitcher": {
+    "title": "Dunklen Modus"
```
fschmenger, line 96:
```suggestion
    "title": "Dunkler Modus"
```

A minor linguistic correction

fschmenger:
Hi @ntma,
thanks for addressing all the proposals that quickly!

> I was trying to fix the hovering on the BgLayerList and ended up with a solution that uses v-hover with a v-overlay. (code here)
It is hard to justify the 0.2 opacity with the color="grey", but it seems to apply a similar overlay to when hovering Vuetify buttons. Is this acceptable for Wegue?

From looking at the wegue code base i can see the reason for hovering not working together with the dark theme that well:
```
v-card--hover:focus, .v-card--hover:hover {
    box-shadow: 0 5px 5px -3px rgb(0 0 0 / 20%), 0 8px 10px 1px rgb(0 0 0 / 14%), 0 3px 14px 2px rgb(0 0 0 / 12%);
```
The dark theme color is already too dark for the box-shadow to take any effect. I think this will hit us in many situations. So I agree with you, that a specific fix at this place is just not worth it. We could either think about making the dark theme default colors just a bit lighter or just ignore the problem at all and see if a future vuetify version has a generic fix for that. Probably just roll it back for now ?!


Regarding the fix introduced for `ToggleButton` in b1e76848ecda65edd0a45ab9b9419ebd61bd95d3.
Doesn't it work much simpler, if you just assign the color property to the `v-btn` in favor of the css rule?
```
<v-btn icon :value="true" @click="toggleUi" :color="color">
    <v-icon :color="color" medium>{{icon}}</v-icon>
</v-btn>
```


Regarding the theming for the Geocoder and the AttributeTableWin, I will look into the solo property stuff and give you feedback later.

fschmenger:
Hi @ntma, I had a look at the `solo` fields now in conjunction with 4c2cbf357bbf25f73331fd156558b87fc39da2ed.

I'm a bit puzzled, why this is actually required. For a test I removed the `solo` attribute, set the `filled` attribute to always true and removed the corresponding css code in wegue.css. Something along these lines, for e.g. `AttributeTable`. 

```
     <v-select
        v-model="selLayer"
        color="accent"
        item-color="secondary"
        :dark="isPrimaryDark"
        filled
        outlined
        class="wgu-vector-layer-select"
        :items="displayedLayers"
		    :item-text="item => item.get('name')"
        :menu-props="{
          bottom: true,
          'offset-y': true,
        }"
        dense
        return-object
        hide-details
        :label="$t('wgu-attributetable.selectorLabel')"
        ></v-select>
```
I have to say I couldn't really spot the difference, but then I didn't test this across all the material studies. Maybe you could review this or give me an explanation. :)

chrismayer:
> Just used a little evening spare time to change the plain html tables of the infoclick module into a vuetify based v-simple-table

@fschmenger thanks for this. Please provide a separate PR for this. I guess @ntma has to merge in the current master from time to time :wink: so I think merging the PR in advance would be an option. The other way round would also be OK, that you place your PR after this (theming) is merged. About the order I have no strong opinion (what fits better for you @fschmenger and @ntma). But any way you choose a separate PR is preferred.

chrismayer:
@ntma @fschmenger @JakobMiksch What is the current stage of this? Does it need more work or is it good to go? No hurry, just for information.

fschmenger:
@chrismayer 
Alright, I will post the changes to infoclick after this as a separate PR, as it depends on theming support.
Regarding the state of this PR, Nuno is almost there. I have a couple of points mentioned above to review on finalization. Also the unit tests are currently failing.

ntma:
Hi @fschmenger @chrismayer, sorry for the absence but I have been very busy with my job. I'll get back to this PR as soon as possible to answer the questions directed to me.

In the meanwhile, merging any PR that depends on this update is ok for me!  I'll rebase the changes later :+1:

JakobMiksch:
thanks for @ntma and @fschmenger for the ongoing work. I did not follow all the recent details anymore. Let me know if you need an additional review. Otherwise I trust the approval of the others.

ntma:
Hi guys, I am back!

So, to reply some of the pending questions/reviews from @fschmenger:

> The dark theme color is already too dark for the box-shadow to take any effect. I think this will hit us in many situations. So I agree with you, that a specific fix at this place is just not worth it. We could either think about making the dark theme default colors just a bit lighter or just ignore the problem at all and see if a future vuetify version has a generic fix for that. Probably just roll it back for now ?!

I think the problem will persist even if we change the dark theme tones, as the hovering shadow will conflict with the dark shades. I could ignore the issue for now but I feel that Vuetify (like many other UI frameworks) are in a lot of pressure to release their stable version for Vue3. Details like this are probably not their priority right now :disappointed: Should I still revert the changes?

> Regarding the fix introduced for ToggleButton in b1e7684.
> Doesn't it work much simpler, if you just assign the color property to the v-btn in favor of the css rule?

Indeed, nice catch! One more caveat for Vuetify :+1: I will change it to your suggestion!

> EDIT: Since the specific UI design of the 2 controls is subject to discussion / change in the future (see #243 and #253) I would rather opt for a simplistic and fairly straightforward - rather than perfect - solution here.

Regarding the solo fields, since we are not opting for the near perfect solution (and I totally agree with this decision), should this field take the same color as the header color?

> I wonder if you like it, it's a little more simplistic. If so I can either provide this as a follow up PR or suggest as a code change and we get it incorporated in this run.

I agree with the infoclick change. Having custom components converted to Vuetify components will simplify the material theming  :+1:

fschmenger:
Hi @ntma,
thanks for the investigation and the feedback.

> I think the problem will persist even if we change the dark theme tones, as the hovering shadow will conflict with the dark shades. I could ignore the issue for now but I feel that Vuetify (like many other UI frameworks) are in a lot of pressure to release their stable version for Vue3. Details like this are probably not their priority right now 😞 Should I still revert the changes?

I'd say yes. Sorry for bringing up the issue at all. But lately I had time to look into it and noticed that this is a general vuetify issue with dark themes. I think we have to live with hovering effects being lost for dark tones for now. IMO having a straightforward code should be prefered over a specific fix, that won't remove the problem at global scope.

> Regarding the solo fields, since we are not opting for the near perfect solution (and I totally agree with this decision), should this field take the same color as the header color?

Did you check the snippet I proposed. I couldn't see the drawback while it looked to me like we're retaining the accent colors. Or am I missing something?

fschmenger:
One last thing from my part, that I cannot see being finally adressed in the discussion above.

> Regarding 4) Are we good to change the background color of the hamburger menu back to 'surface color' ? This is probably a bit opionated, as the hamburger menu is a special dropdown which can also be seen as an extension to the app-bar (and also requires some coloring switches in the code). However I can see the point of Nuno's recommendation.

Again i can only point out, that for my personal taste I would opt for the simplest code-wise solution, and use the primary color for the hamburger menu (especially as it seems to me as this is not explicitly discouraged by the MD spec). I propose @chrismayer should make a clear ruling on this and then we go with whatever he has to say.

ntma:
Hi @fschmenger ,

> I'd say yes. Sorry for bringing up the issue at all. But lately I had time to look into it and noticed that this is a general vuetify issue with dark themes. I think we have to live with hovering effects being lost for dark tones for now. IMO having a straightforward code should be prefered over a specific fix, that won't remove the problem at global scope.

It's totally fine. I understand that reaching a future proof solution isn't always easy :slightly_smiling_face: I'll revert the changes in the next commit.

> Did you check the snippet I proposed. I couldn't see the drawback while it looked to me like we're retaining the accent colors. Or am I missing something?

I did. The main difference is the 3rd elevation added by the 3rd dark shade (docs about  material elevation [here](https://material.io/design/color/dark-theme.html#properties)).  Here are two visuals (first with solo and second without), where the red numbers represent the three elevations.

##### With solo:
solo_numbered 

##### Without solo:
no_solo_numbered

Elevation is essential to differentiate components (specially in the dark mode). Lighter shades will simulate positive elevation while darker shades will simulate negative.

In my point of view, solo fields should that a negative elevation (in respect to their parent components). Although they are components that should stand out, they are at the same time recipients for user input. And thus should look like an embedded field into it's parent component. Hence my effort on trying to make it look like that :grin:

fschmenger:
Hi @ntma, thanks for the screenshots regarding the solo fields - I can spot the difference now :)

I'm a bit undecided whether this is a good feature for the following reasons:
* Wegue didn't have any concept on accecting those fields over a primary surface prior to this PR (I think this has been mentioned somewhere above)
* The implementation is currently limited to 2 controls, that may be subject to changes in the future.
* Last but not least the solution is not very maintainable. If we we're to add future controls over a primary surface, then these would require the same code switches and most importantly the same css rules that you introduced in wegue.css.

Overall to me the feature looks too minor to justify a bunch of specific code.
I leave the final decision to the meggsimum guys - @chrismayer, @JakobMiksch what's your opinion on this?

IMO, if we decide to keep this in, then for the sake of maintainance you should probably wrap up the wegue.css rules by a class and assign it to the respective controls.

chrismayer:
Hi @ntma,

thanks for your ongoing work and your gigantic patience with the huge amount of remarks, discussions and suggestions by the reviewers. I really appreciate that you're still on this :+1: But to not overstretch you patience :wink:  and to get this killer feature into the Wegue code base we should shorten this a bit up. Therefore I propose the following: 

We should go the simplest approach for open remarks/request to get them quickly done. For the hamburger menu we should use using primary color (simplest approach? Correct me if I am wrong). 
For solo fields I have no strong opinion, please also go a simple, fast and acceptable way here.
All other (hopefully minor) things, which can be improved in the future we should live with for now. These should be formulated in issues, so they are not forgotten.

So @ntma would you please do a last round of providing final adaptions, if necessary? Would be very appreciated. Afterwards give us a ping that this is finalized. 
@fschmenger Would you give afterwards a last check regarding heavy bugs or blockers? If none I'd happily merge this and we proceed from there.

Again thank you all (especially @ntma and @fschmenger) for your huge work on this!

ntma:
Hi @chrismayer ,

Glad that you appreciate the effort done in this PR! :grin:

> So @ntma would you please do a last round of providing final adaptions, if necessary? Would be very appreciated. Afterwards give us a ping that this is finalized.

I did the last round and:

* the hamburger menu is now primary;
* `solo` fields are now `filled` fields. The visual difference can be seen in  [this comment](https://github.com/meggsimum/wegue/pull/262#issuecomment-992700598). In the end, having a 3rd elevation may not look that bad **as long as this property is uniform across all fields over primary colors**. Unfortunately [this nasty override](https://github.com/meggsimum/wegue/pull/262/files#diff-4195023791faf7abae8098f5e734fa768c867aedaeb0d3b55510ab391c2ad91aR52) is still required for the cases where we override the "onprimary" color (ie: app-conf-projected.json).

As a last remark, this PR adds the ability of using the Vuetify color classes as CSS variables (docs [here](https://vuetifyjs.com/en/features/theme/#custom-properties)). I didn't test it but I believe these variables can be used to style the map features if required.

And with that done, I conclude this work. Thank you all for the patience on reviewing this PR. If there is still something that I missed or requires changes, please let me know :+1:

fschmenger:
Hi @ntma, thanks for your ongoing effort.

I can see 2 finalization tasks:
* In conjunction with e8b4760042348fe21c2280e12ea269c6ad97ba3b: Could you please remove the following line https://github.com/meggsimum/wegue/blob/43eb1375c9251aeca9f33cc280863c32b5e87302/app-starter/components/AppHeader.vue#L112 and the associated color props from the components/ buttons, so we are consistent with hardwiring of colors on component level.
* Are the specific css rules still required since 43eb1375c9251aeca9f33cc280863c32b5e87302? Please either remove them or see my comment above for attaching them to a css class (e.g. 'wgu-select'). https://github.com/meggsimum/wegue/blob/43eb1375c9251aeca9f33cc280863c32b5e87302/src/assets/css/wegue.css#L42-L62

Otherwise this looks good to me!

ntma:
Hi @fschmenger,

> In conjunction with e8b4760: Could you please remove the following line and the associated color props from the components/ buttons, so we are consistent with hardwiring of colors on component level.

Makes sense. Done!

> Are the specific css rules still required since 43eb137? Please either remove them or see my comment above for attaching them to a css class (e.g. 'wgu-select'). 

Unfortunately yes. As far as I know, there is no simple way of overriding the text/border/input color in case the field is over a primary color. The CSS overrides are now wrapped into `.wgu-solo-field`.

Cheers :+1:

fschmenger:
Hi @ntma, thank you for the prompt response and fixes.

> Unfortunately yes. As far as I know, there is no simple way of overriding the text/border/input color in case the field is over a primary color. The CSS overrides are now wrapped into .wgu-solo-field.

I'm a little bit confused again, as I can't see the rule taking effect now - why the name of the css class as the associated components are now no longer `solo`? I asked my DOM inspector to point me to the difference but never saw the `theme--light` assigned (the `wgu-solo-field`s always have `theme--dark` in the class list, even when I toggle the dark mode).
Maybe you can point me to the difference again?

fschmenger:
@chrismayer I reviewed the code and ran some additional integration tests on my app with it's specific theme. I think this PR is as good as it can get, so feel free to merge.

@ntma Thanks a lot for going through all the proposals and all the time and energy you invested into this. I'm looking forward to work with you on some topic again in the future! 👍

JakobMiksch (approved):
Fantastic work. 
I tried it again and I love it! 
I also tried the built version produced by `npm run build` and it seems to work well. 
Ready to merge from my side!

chrismayer:
**THIS IS GREAT!** :rocket: 

Thanks for your fantastic and tireless work @ntma. Very much appreciated. I hope you  stay part of the Wegue community. Would be a great benefit for us all.

Also big thanks to @fschmenger for managing this and for the endless reviews. And of course thanks to @JakobMiksch for your input.

I am going to merge now. :tada:

ntma:
> Thanks for your fantastic and tireless work @ntma. Very much appreciated. I hope you stay part of the Wegue community. Would be a great benefit for us all.

Anytime! I'll try to contribute more whenever I find myself with some spare time :+1:

PR #264, opened 2021-11-09
Permalink layer exclusion

### LINKED ISSUE #244: Exclude temporary layers from permalink
The permalink module currently registers for changes to all layers in the OL layer list and incorporates the state of all layers in it's link URL. As described in #240 we should mostly ignore non application wide layers, particularly those added on a module level since those won't work unless the state of the module is restored too.

A possible solution could look like this:
* The layer factory will add a custom attribute 'supportsPermalink' to the OL layer list and default this flag to true (maybe with an option turn it off via config).
* Permalink controller evaluates this flag to enable/disable layers for inclusion.
* Static ignore list in permalink controller can be removed

By default we will then have permalink support for application wide layers but not dynamical layers added by application code. However the latter can still decide, whether it makes sense to flag its own layers as usable for permalink and set the attribute 'supportsPermalink' on layer creation.

fschmenger:
This branch implements dynmical exclusion of layers from permalinks by means of a newly introduced layer attribute `supportsPermalink`. The corresponding issue is described in #244 (resolves #244).

I also added a documentation section for the `permalink` configuration, which did not exist yet.

### DIFF
--- a/docs/map-layer-configuration.md
+++ b/docs/map-layer-configuration.md
@@ -17,7 +17,7 @@ The following properties can be applied to all map layer types
 | opacity            | Numeric value ranging from 0 to 1 describing the opaqueness of the layer. Defaults to `1.0`. | `"opacity": 0.5` |
 | zIndex             | Numeric value specifying the stack order of layers. Layers will be ordered by z-index and then by order of declaration. Defaults to `-1` for background layers and `0` for all other layers.  | `"zIndex": 2` |
 | displayInLayerList | Boolean value, whether the layer should appear in the LayerList. Ignored if the layer is a background layer - see option `isBaseLayer`  | `"displayInLayerList": true` |
-| supportsPermalink  | Boolean value, whether the layers state should be considered   in permanent links - see also [permalink](wegue-configuration?id=permalink). Defaults to `true`.  | `"supportsPermalink": true` |
+| supportsPermalink  | Boolean value, whether the layers state should be considered in permanent links - see also [permalink](wegue-configuration?id=permalink). Defaults to `true`.  | `"supportsPermalink": true` |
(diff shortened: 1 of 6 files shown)


fschmenger:
Maybe someone more familiar than me with the `PermalinkController` could review the newly added configuration section in the docs.

chrismayer:
@justb4 do you want to have a look since the `Permalink` implementation was by you initially? Thanks in advance.

docs/map-layer-configuration.md
```diff
@@ -17,6 +17,7 @@ The following properties can be applied to all map layer types
 | opacity            | Numeric value ranging from 0 to 1 describing the opaqueness of the layer. Defaults to `1.0`. | `"opacity": 0.5` |
 | zIndex             | Numeric value specifying the stack order of layers. Layers will be ordered by z-index and then by order of declaration. Defaults to `-1` for background layers and `0` for all other layers.  | `"zIndex": 2` |
 | displayInLayerList | Boolean value, whether the layer should appear in the LayerList. Ignored if the layer is a background layer - see option `isBaseLayer`  | `"displayInLayerList": true` |
+| supportsPermalink  | Boolean value, whether the layers state should be considered   in permanent links - see also [permalink](wegue-configuration?id=permalink). Defaults to `true`.  | `"supportsPermalink": true` |
```
JakobMiksch, line 20:
```suggestion
| supportsPermalink  | Boolean value, whether the layers state should be considered in permanent links - see also [permalink](wegue-configuration?id=permalink). Defaults to `true`.  | `"supportsPermalink": true` |
```

JakobMiksch (approved):
I tried it and it works like expected. 
Can be merged from my side.

fschmenger:
@JakobMiksch Thanks for the review, Jakob!

fschmenger:
This PR has been rebased to the current master to resolve documentation conflicts, there should be no changes. IMO we're good to merge this, but a quick review definitely can't hurt. :)

JakobMiksch:
I gave it a quick test and it seems to be fine :+1:
Ready to merg!

PR #267, opened 2021-11-25
Use new PDOK background layers in projected example app

### LINKED ISSUE #226: Remove no longer working layers from the configuration examples.
Ahocevar-wms (https://ahocevar.com/geoserver/wms) and Vector Tile Layer (https://ahocevar.com/geoserver/gwc/service/tms/1.0.0/ne:ne_10m_admin_0_countries@EPSG%3A900913@pbf/{z}/{x}/{-y}.pbf) seem to be no longer working:
Any request results in 502 -Bad Gateway.

I discovered this today while testing layer sources for being CORS enabled.

This could be a temporary problem. If it persists we should remove the layers from our configuration templates.

chrismayer:
This replaces the old (no more existing) PDOK background layers (topo map and aerial image) by their official replacements.

Fixes #226


justb4:
Thanks @chrismayer ! Looks good to me.

chrismayer:
Thanks for the review and the info about the new map services @justb4 :+1:

PR #268, opened 2021-12-03
eslint: weaken rules in dev mode

JakobMiksch:
- weaken eslint rules in development mode
- [ ] maybe use with different command than `npm run dev`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,12 +18,16 @@ module.exports = {
   // add your custom rules here
   'rules': {
     // allow paren-less arrow functions
-    'arrow-parens': 0,
+    'arrow-parens': 'off',
     // allow async-await
-    'generator-star-spacing': 0,
+    'generator-star-spacing': 'off',
     // allow debugger during development
-    'no-debugger': process.env.NODE_ENV === 'production' ? 2 : 0,
+    'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
     // allow semicolons and the end of a statement
-    'semi': 0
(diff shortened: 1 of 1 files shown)


chrismayer (approved):
LGTM, thnaks for this. I would not introduce another npm task. Having this in the `npm run dev` should enough.

fschmenger:
I like the idea of being a little more fault tolerant in dev mode, since currently this is a pain to get all the formatting right when we're just testing some code. 
On the other hand, I don't like to see the production build failing when I haven't been warned about it dev mode. What do you think about turning all the disabled rules into warnings instead?

chrismayer:
> What do you think about turning all the disabled rules into warnings instead?

I think that is exactly what @JakobMiksch did.

fschmenger:
Ok, thanks for the info! I didn't test it but from the code it looked to me like some errors are now generally disabled for dev mode (`arrow-parens`, `no-debugger`, ...)
Here's a possible generic approach to do it with a plugin - maybe something along these lines is interesting to switch back and forth between warning levels without touching any eslint configs?
https://stackoverflow.com/questions/44303590/set-default-error-level-to-warn

JakobMiksch:
@fschmenger This change only applies when running in `dev` mode. When eslint is unhappy it "only" shows warnings both in the terminal and the browser console, but does not break the application instantly. Hovever, when you run `npm run build` or `npm run test` the remarks will be errors. 

The idea was to improve the developer experience and make development a bit less laborious. Especially when you make a quick prototype of something ...

PR #272, opened 2022-01-11
Add a sample module

### LINKED ISSUE #265: Describe how to add a custom module to Wegue
I got the hint that there is no sufficient description for new users how to include a custom module into the application. 
- a note in the docs would be useful
- an example module in the `app` directory could help new users to get started easier

JakobMiksch:
fixes #265 

- This PR adds a minimal module to `app-starter` that should help new users to create their custom modules. 
- This new `SampleModule.vue` is extensively commented to explain the basic anatomy of a module
- It contains a simple access of the OpenLayers map to make creating custom map-related modules easier

image

### DIFF
--- a/app-starter/WguAppTemplate.vue
+++ b/app-starter/WguAppTemplate.vue
@@ -81,6 +81,7 @@
   import MapLoadingStatus from '../src/components/progress/MapLoadingStatus'
   import AttributeTableWin from '../src/components/attributeTable/AttributeTableWin.vue'
   import MapRecorderWin from '../src/components/maprecorder/MapRecorderWin'
+  import SampleModuleWin from './components/SampleModule.vue'
 
   export default {
     name: 'wgu-app-tpl',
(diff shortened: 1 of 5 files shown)


fschmenger:
Hi @JakobMiksch, this a first good step in the right direction!

Before you close #265, keep in mind, that we probably need some sort of API documentation for the module-cards. The available generic properties are at least documented on the app-conf level, so maybe this isn't urgent yet.
Also my point from the comment still stands, that we should factor out / separate the examples from the starter templates on the long run.
Maybe you could set up some further issues (1. reusable component doc, 2. app-starter example separation) from that or just keep #265 open?!

JakobMiksch:
@fschmenger thanks for you comment. I agree with you that we should not forget about the other parts of #265

JakobMiksch:
I created two new issue #274 and #275 
Then #265 can be safely closed

chrismayer (approved):
Thanks for this. Please merge.

PR #277, opened 2022-01-18
Minor fix for overview map

fschmenger:
Fixes a scrollbar bug for the overview map reported by @JakobMiksch.


chrismayer (approved):
Thanks again @fschmenger

JakobMiksch (approved):
@fschmenger thanks for the fix. It can be merged

PR #278, opened 2022-01-18
Infoclick vuetify tables

fschmenger:
This pull request changes the plain html tables of the infoclick module into a vuetify based v-simple-table. The idea is to get rid of a lot of custom css formatting and use a control that is already theming compliant. 
This is a follow up task to #262 and was previously mentioned there.

Here's a screenshot of the result:
wegue_map_click_info

### DIFF
--- a/src/components/infoclick/CoordsTable.vue
+++ b/src/components/infoclick/CoordsTable.vue
@@ -1,31 +1,19 @@
 <template>
 
-  <table 
-    :class="{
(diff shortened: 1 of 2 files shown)


JakobMiksch:
I tried it and it works well. Nice addition. 
 Ready to merge from my side.

chrismayer:
Very nice, thanks @fschmenger. Going to merge now.

PR #279, opened 2022-01-24
Replace fetch by axios

fschmenger:
This is a proposed cleanup task by @JakobMiksch from #269. The Fetch API is replaced by axios, no functional changes.


JakobMiksch (approved):
Thanks for the fix. I will merge.

Issue #252, opened 2021-09-07
Improve box model for floating windows.


fschmenger:
Currently, all floating windows (module cards) are descendants of `<v-app>` as implemented in `WguAppTemplate.vue`. Thus all windows are relative to the vuetify root container. This results in a couple of problems:

* Positioning css / config settings has to add offsets, so the windows are not overlapped by the toolbar or footer.
* Mixed configurations with sidebar and floating windows don't go well together - as the opened sidebar may overlap floating windows. 
  * A special case of this is the ` AttributeTable`, which always takes full parent container width and therefore will be partially hidden by the sidebar.
* If the window is draggable, then points 1 and 2 apply, since the window can be dragged under the toolbar, footer or sidebar.

Proposed solution:
Make floating windows descendants of the `<v-content app>` app container. Here's a snippet:

```
<v-content app>
      <v-container id="ol-map-container" fluid fill-height class="pa-0"> .... </v-container>
      <!-- Initialize floating windows here: --->
       <template v-for="(moduleWin, index) in floatingWins">
        <component
          :is="moduleWin.type" :key="index" :color="baseColor"
          v-bind="moduleWin"
        />
    </template>
</v-content>
```

Expected side effects:
* Floating windows will be shifted by the width of the sidebar, once the sidebar is opened / closed. 
  * @chrismayer @JakobMiksch Maybe you could have a word on this, whether this behavior would be ok for you!? 
* Positioning of the windows in app-conf and css code have to be altered to obtain same results. 

This issue is somewhat linked to #215 and could be reviewed in the same run.

PR #292, opened 2022-04-25
Tweaks for DraggableWin directive

ntma:
Fixes:
* [x] Support for Firefox (e.path is Google Chrome specific);
* [x] Reset dragger offset for each mousedown event (prevents magical jumps);
* ~Constain floating windows to the main container;~ (future PR)

#252

### DIFF
--- a/src/directives/DraggableWin.js
+++ b/src/directives/DraggableWin.js
@@ -116,7 +116,7 @@ export const DraggableWin = {
         return;
       }
 
-      // set the offset for each moousedown
+      // set the offset for each mousedown
       setDraggerOffset(el, _data);
 
       if (_data.overlay) {
--- a/src/directives/DraggableWin.js
+++ b/src/directives/DraggableWin.js
@@ -46,7 +46,6 @@ export const DraggableWin = {
     elmnt.addEventListener('mouseup', (e) => DraggableWin.mouseup(e, elmnt, dragConfig));
     elmnt.addEventListener('mousedown', (e) => DraggableWin.mousedown(e, elmnt, dragConfig));
     elmnt.addEventListener('mousemove', (e) => DraggableWin.mousemove(e, elmnt, dragConfig));
-    setDraggerOffset(elmnt, dragConfig);
     dragConfig.initialZIndex = elmnt.style.zIndex;
 
     /**
@@ -108,10 +107,17 @@ export const DraggableWin = {
       // id="drag-header" or class="drag-header" should be draggable.
       // If the user clicked another area, do nothing.
       // By default it is restricted to class="wgu-win-title"
+
+      // get the element path
+      const path = e.path || (e.composedPath && e.composedPath());
+
       if (_data.draggableElementSelector &&
-          !checkIfSelectorInPath(_data.draggableElementSelector, e.path)) {
+          !checkIfSelectorInPath(_data.draggableElementSelector, path)) {
         return;
       }
+
+      // set the offset for each moousedown
+      setDraggerOffset(el, _data);
 
       if (_data.overlay) {
         _data.overlay.remove();


fschmenger:
@ntma Hi Nuno, thanks for kickstarting this! I'm currently a little limited in time but can give it a test run by next week hopefully. 

Keep in mind that Task 3) will be a breaking change due to the changes in positioning offsets. Probably this should be in a separate PR, so it can be incorporated with the next Wegue major version.

ntma:
Hi @fschmenger , no problem!

> Keep in mind that Task 3) will be a breaking change due to the changes in positioning offsets. Probably this should be in a separate PR, so it can be incorporated with the next Wegue major version.

Indeed, I got  to agree with that. I will take your suggestion!

src/directives/DraggableWin.js
```diff
@@ -108,11 +107,18 @@ export const DraggableWin = {
...
+      // get the element path
+      const path = e.path || (e.composedPath && e.composedPath());
+
       if (_data.draggableElementSelector &&
-          !checkIfSelectorInPath(_data.draggableElementSelector, e.path)) {
+          !checkIfSelectorInPath(_data.draggableElementSelector, path)) {
         return;
       }
 
+      // set the offset for each moousedown
```
JakobMiksch, line 119:
```suggestion
      // set the offset for each mousedown
```

JakobMiksch (commented):
I do not really understand the change here. Could you explain what the behavior before and after this PR is?

src/directives/DraggableWin.js
ntma: Hi @JakobMiksch,

To reproduce the issue, open two windows configured with `draggable: true` and drag one window after the other

**Before this PR**, the current window being dragged will always offset to the position of the previously dragged window.

**After this PR**, the offset is resetted on each drag start.

This issue is caused by the following object being shared across all components that use the `DraggableWin.js` directive.
https://github.com/wegue-oss/wegue/blob/7f53a55545015d37c5b4fc73c8f9dba334240b7d/src/directives/DraggableWin.js#L9-L26

JakobMiksch: Alright. I have actually never used this dragging functionality so far. 
I only could drag the InfoClick window so far. The other ones do not work to drag. 
Do you know why, which windows do you use for testing?

ntma: I tested with `wgu-maprecorder` and the `wgu-infoclick` in google chrome. Had to set `draggable: true` to both of them in the app-conf.json.  Are you able to reproduce?

JakobMiksch: Yes, I was able to reproduce. Good catch.

JakobMiksch (approved):
Looks fine from my side. I tested it and it does what it should.

JakobMiksch:
@fschmenger maybe could have a look and merge if you approve.

src/directives/DraggableWin.js
chrismayer: Other windows like the measure window do not get draggable even with the `draggable: true` in the `app-conf.json`. Reason is the positioning in the projects CSS file, where already a TODO is noted, that the positioning has to be harmonized, see https://github.com/wegue-oss/wegue/blob/master/src/assets/css/wegue.css#L112

chrismayer:
Thanks @ntma for this. LGTM :+1: As @JakobMiksch suggested, let's see what @fschmenger says and if he approves feel free to merge afterwards.

fschmenger:
OK, this looks good to merge. I can finally see the `draggable` option working in Firefox and it no longer get's confused when dragging multiple cards. 

I suggest we leave #252 open, until the other topics can be addressed for v2.

PR #295, opened 2022-07-21
Upgrade non-dev dependencies

chrismayer:
This upgrades the "non-dev" dependencies to minor and patch releases, which can applied without significant code changes.

OpenLayers upgrade should be done separately since there has been some work done in #261


fschmenger:
Hi @chrismayer,
I finally found some time to give this a brief test. At the time of writing this some package (minor) versions no longer correspond to the newest ones available, so I decided to give it a shot with this configuration:

```
"dependencies": {
    "axios": "0.27.2",
    "canvas-record": "3.1.0",
    "echarts": "^5.3.2",
    "ol": "6.4.3",
    "proj4": "2.8.0",
    "vue": "2.7.13",
    "vue-echarts": "^6.0.0",
    "vue-i18n": "8.28.2",
    "vuetify": "2.6.12"
  },
 "devDependencies": {
    ...
    "vue-template-compiler": "2.7.13",
}
```

Here are my findings so far:
* A warning appears: `[Vuetify] [UPGRADE] 'v-content' is deprecated, use 'v-main' instead.`
* In my custom application it breaks some internal dependencies of apache-echarts (`"vue-echarts": "^6.0.0"`), so these are no longer working.
* My hover-cards changed in layout and are now somehow truncated at the bottom (no special css is involved there).
* As you already discovered, the css rule regarding the app logo has to be changed - apparently there are some internal changes to classes assigned by vuetify.

This might be just the tip of the iceberg. So after all my conclusion is, that we should do this in the same run as #297 (or the upcoming cli-v4-PR)  when moving to Wegue v2.

Greets, Felix.

sronveaux:
Hi @chrismayer ,

I just wanted to complement a little bit the comment of @fschmenger.

I personally think you should indeed wait to be using at least the VueCLI V4 for the following two reasons:

- If you read the [upgrade guide to Vue 2.7](https://github.com/vuejs/vue/blob/main/CHANGELOG.md#upgrade-guide), it is written that you first have to upgrade your local `@vue/cli-xxx` dependencies to the latest version in your major version range... and only v4 and v5 are listed there.
It doesn't mean that it can't work with Vue-CLI V2 which is used now in Wegue, just that it is not supported...
- Concerning `Vuetify` and the warnings, they have developed a plugin for `ESLint` which returns all those deprecation warnings. However, this plugin needs `ESLint6` at minimum and wegue currently uses `ESLint5`. The upgrade of VueCLI also upgrades `ESLint`. So we'll be able to add the `Vuetify` plugin in Wegue to help find those potential problems.
I once read a little bit what was new and it unfortunately seems this is the only mentioned big change to do (replacing `v-content` by `v-main`). This means all the display errors that @fschmenger discovered will certainly have to be tackled one by one unfortunately...

Cheers,
Sebastien

spwoodcock:
Would it be helpful for me to merge this into #320, to consolidate the changes and resolve conflicts?

chrismayer:
Superseded by #335 - closing.

PR #298, opened 2022-08-01
Update to last Vue CLI v2 template

sronveaux:
As discussed with @chrismayer by mail a few months ago, I tried to upgrade Wegue to newer versions of Vue-CLI and Webpack (just to mention those two). See https://github.com/wegue-oss/wegue/issues/94.

This project was going well until the end of April when I unfortunately had no more time to work on this. Here I had a little time again and managed to get almost to the end of it. So I think it's time to share...

I'll make a serie of PR, each going to the latest template used by the next Vue-CLI version.

For this one, nothing really big, it's just going to the latest template of Vue CLI V2. The major improvements are linked to a fully functional ESLint and an upgrade from Webpack 2.7 to Webpack 3.6.

This is also using Webpack-dev-server instead of running Express directly.
There is a bug here though in this version of the dev-server (which will be resolved in next PR with Vue-CLI V3) where compilation warnings issued while in dev mode are all reported as errors. So you can have the browser overlay report everything as errors or hide it completely (which I have done here). You can easily put it back on by going in the [build/webpack.dev.conf.j](https://github.com/sronveaux/wegue/blob/lastVueCLI2/build/webpack.dev.conf.js#L38)s file and change the overlay parameter, setting warnings to true instead of false.

I hope to post a PR for Vue V3 this afternoon, V4 should follow shortly and I still have to work a little bit on V5.

If you have ideas or requests to configure some things differently, don't hesitate to ask, I'll try to add what I can in coming PRs.


JakobMiksch:
thanks @sronveaux for the PRs, I am starting to review. One question right away. 
Why are there individual PRs for each version of Vue-CLI?
I guess finally, we would only like to use one of them (probably, the latest), right?

JakobMiksch:
@sronveaux I ran Wegue with this setup and it works like expected great. 
This PR makes a lot of formatting of the source code, which makes the PR challenging to review. 
Is this code formatting required by performing the update to Vue-CLI v2?

sronveaux:
@JakobMiksch thanks for the reply and the comments, that's good news to see that it works as expected !

> thanks @sronveaux for the PRs, I am starting to review. One question right away. Why are there individual PRs for each version of Vue-CLI? I guess finally, we would only like to use one of them (probably, the latest), right?

The individual PRs are in fact cumulative. So for the code to be fully upgraded, you must apply this PR then the second one (#297).
The idea was to give you the opportunity to test and stop upgrading at the point where you think nothing interesting is added.

> @sronveaux I ran Wegue with this setup and it works like expected great. This PR makes a lot of formatting of the source code, which makes the PR challenging to review. Is this code formatting required by performing the update to Vue-CLI v2?

Concerning the code formatting, there are two different reasons for that which you can find if you follow the GIT logs:

1. Some are there as the VUE template code was updated that way (var changed to const for example). The idea was to stay in tune as much as possible with the boilerplate code that comes with VUE.
2. The commit 77b0b754ca1331ea8b3c8fbd0f8baa407deda312 is not needed to upgrade correctly but makes a lot of changes. This is the result of running eslint and fixing the reported errors. There currently is a bug in Wegue configuration and because of it, eslint is not running inside .vue files. This commit is the result of running eslint on .vue files now that it works as expected.

One last important thing is the package-lock.json file. Commit abbf7cca958640b0e5eec7d6525b8017f9a5a031 changes more than 30.000 files alone. It is if fact just a reset to the state before applying all the other commits. This was done to avoid some dependencies problems and is linked to the first comment of @fschmenger on #294. package-lock.json has to stay "outdated" for now as you start getting into trouble after the first upgrade to a newer file format. This is linked to Webpack V2.7 being used by Wegue when some of its dependencies are requiring Webpack V4. Because of this, newer versions of npm force you to use --legacy-peer-deps in order to install things correctly. This is solved by PR #297
 
Hope this makes sense, don't hesitate if you have more questions!

PR #300, opened 2022-09-23
Allow rendering of HTML in PropertyTable

chrismayer:
This allows the rendering of HTML in the value column of the `PropertyTable`. Normal (non-html) text is rendered as before, so we have backward compatibility here. No need for updates of existing apps.


JakobMiksch (approved):
I tried it and it works like expected. Feel free to merge.

chrismayer:
Thanks for the review @JakobMiksch! I am going to merge now.

PR #302, opened 2022-10-06
Add event 'sidebar-toggle' to programmatically toggle sidebar visibility

### LINKED ISSUE #301: Hook to toggle sidebar visibility programmatically
I'd like to have a common hook to toggle the visibility of the app sidebar programmatically (not by a user click). I have a need for that in several Wegue applications and maybe there is chance to add this into the Wegue core functionality. I see differerent possibilities to achieve that:

**By event**

One possibility would be to define an app-wide event like `'wgu-sidebar-toggle'`, on which the `AppSidebar.vue` listens to:

    WguEventBus.$on('wgu-sidebar-toggle', () => {
      this.toggleOpen();
    });

Once any custom app code emits this event (`WguEventBus.$emit('wgu-sidebar-toggle');`) the visibility of the sidebar would be toggled. 

**By ref**

Another possibility would be to add a ref for the `</wgu-app-sidebar ref="wgu-app-sidebar">` in the `WguAppTemplate.vue`. Then at least in the `WguApp.vue` of the app the sidebar would be accessible via `this.$refs['wgu-app-sidebar']` and could be toggled by `this.$refs['wgu-app-sidebar'].toggleOpen()`. But here I see the problem that the `WguAppTemplate.vue` of existing apps needs to be updated. 

Since @fschmenger is the architect of the sidebar concept I really would appreciate his opinion on the best approach from your point of view (maybe another besides the 2 above) or if there is already a possibility for that, which I missed. Thanks in advance.

chrismayer:
This adds a hook to programmatically toggle the open state of the `AppSidebar`. It is implemented as event `'sidebar-toggle'`, on which the `AppSidebar` listens to. So every emit of 'sidebar-toggle' over the WguEventBus (`WguEventBus.$emit('sidebar-toggle');`) toggles the visibility of the sidebar.

Resolves #301

### DIFF
--- a/app-starter/components/AppSidebar.vue
+++ b/app-starter/components/AppSidebar.vue
@@ -47,6 +47,9 @@ export default {
     WguEventBus.$on('sidebar-scroll', comp => {
       this.scrollTo(comp);
     });
+    WguEventBus.$on('sidebar-toggle', () => {
+      this.sidebarOpen = !this.sidebarOpen;
+    });
   },
   methods: {
     /**


fschmenger (approved):
LGTM!

chrismayer:
Thanks for the review, going to merge now.

PR #306, opened 2022-12-06
Additional HTTP params for WMS layers

chrismayer:
This adds the optional config option `params` for `'TILEWMS' `and `'IMAGEWMS'` layers. This allows to inject custom HTTP parameters to the GetMap request of the layer.

This is useful to set custom (non-standardized) WMS request parameters, such as `CQL_FILTER` or `FEATUREID`, which are e.g. supported by the GeoServer WMS implementation or the `angle` parameter of the UMN Mapserver software.


fschmenger:
@chrismayer: 
Nice job! This has also been on my TODO list for quite some time. From looking at the code this looks good, but didnt have any time to give it a test run yet. 
imo for the sake of simplicity you could just name the new attribute `params`, but that`s a matter of taste.

chrismayer:
Thanks for your review @fschmenger . I agree that using `params` for simplicity would be better. I adapted my changeset in [e2a7a62](https://github.com/wegue-oss/wegue/pull/306/commits/e2a7a621aa924c72744b509aa487efcedc2143f5) . Would you please have another short look and approve the PR so I can merge? Thanks in advance.

fschmenger (approved):
Ok, just gave the parametrized TILEWMS a test run and it works out as expected. Can be merged.

chrismayer:
Great! Thanks again @fschmenger!

PR #307, opened 2022-12-09
Fix typo in lid for vector tiles layer

chrismayer:
Minor typo correction within the `lid` of the vector tiles layer in the `app-starter` package and update of the corresponding docs. No functional change.


JakobMiksch (approved):
Good catch.

chrismayer:
Thanks for the review @JakobMiksch

PR #308, opened 2022-12-09
Add app-wide loading mask component

chrismayer:
This adds an app-wide loading mask component, which covers the application and shows a loading spinner:

image

Quite useful when performing essential AJAX requests, where the user should not interact with the application.

To toggle the loading mask you could use `WguEventBus.$emit('app-loading-mask-toggle');` or in order to explicitly set a visibility for the mask one could use `WguEventBus.$emit('app-loading-mask-toggle', newViz);` .

### DIFF
--- a/app-starter/WguAppTemplate.vue
+++ b/app-starter/WguAppTemplate.vue
@@ -58,6 +58,10 @@
 
     <slot name="wgu-after-footer" />
 
+    <!-- app wide loading mask,
+    use WguEventBus.$emit('app-loading-mask-toggle', newViz) to show/hide -->
+    <wgu-app-loading-mask />
+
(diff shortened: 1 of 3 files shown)


fschmenger (approved):
This works out as expected. Another useful feature. 
You could possibly enhance the unit test to `$emit` the event and check the `show` attribute.
Approved anyway, if you want to merge.

Since we have a couple of new features by now, could you bump up the minor version?

chrismayer:
Thanks for your review @fschmenger. I added some tests regarding the 'app-loading-mask-toggle' event (see 10d0875). I am going to merge now.

> Since we have a couple of new features by now, could you bump up the minor version?

I also had that in mind and come up with that separately.

PR #313, opened 2022-12-21
Resizable sidebar

fschmenger:
This pull request adds additional options to the sidebar configuration to make it resizable on behalf of the user - a resizer appears when hovering the right edge of the sidebar.

Enable the option in app.conf (defaults to false):
```
"sidebar": {
    "resizable": true
}
```

Set additional resizing constraints (no restrictions per default)
```
"sidebar": {
    "minWidth": 400,
    "maxWidth": 600
}
```

A technical note:
To resize the OL map canvas automatically when it`s container size changes, I hooked up a [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) in Map component. This should have sufficient browser support by now. It also fires initially so the workaround on application start has been removed.
This also means we can remove other map-resize related code like in attributeTable. I will provide those with a couple of fixes in an upcoming PR.

Merry christmas to all of you!

### DIFF
--- a/app-starter/components/AppSidebar.vue
+++ b/app-starter/components/AppSidebar.vue
@@ -78,22 +78,21 @@ export default {
      * Resize the sidebar, if the 'resizable' option is enabled.
      */
     onResize () {
-      const me = this;
-      const sidebarEl = me.$refs.sidebar.$el;
+      const sidebarEl = this.$refs.sidebar.$el;
       sidebarEl.style.transition = 'initial';
 
       // Resize on mouse move
-      function onMouseMove (e) {
+      const onMouseMove = (e) => {
         document.body.style.cursor = 'ew-resize';
         let w = e.clientX;
-        w = Number.isNaN(me.minWidth) ? w : Math.max(me.minWidth, w);
-        w = Number.isNaN(me.maxWidth) ? w : Math.min(me.maxWidth, w);
-        me.sidebarWidth = w;
+        w = Number.isNaN(this.minWidth) ? w : Math.max(this.minWidth, w);
+        w = Number.isNaN(this.maxWidth) ? w : Math.min(this.maxWidth, w);
+        this.sidebarWidth = w;
       }
       document.addEventListener('mousemove', onMouseMove, false);
 
       // Stop the interaction on the next mouse up.
-      function onMouseUp () {
+      const onMouseUp = () => {
         sidebarEl.style.transition = '';
         document.body.style.cursor = '';
         document.removeEventListener('mousemove', onMouseMove, false);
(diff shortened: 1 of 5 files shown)


src/components/ol/Map.vue
```diff
@@ -65,7 +65,18 @@ export default {
     // already been fired. Don not use directly in cmps, use Mapable instead.
     Vue.prototype.$map = me.map;
 
-    me.map.setTarget(document.getElementById('ol-map-container'));
+    // Set the target for the OL canvas and tie it`s size to ol-map-container.
+    // Remarks: 'ol-map-container' does not exist in the scope of the current unit test,
+    //  therefore flag the initialization to prevent errors.
```
JakobMiksch, line 70:
```suggestion
    // therefore flag the initialization to prevent errors.
```

app-starter/components/AppSidebar.vue
```diff
@@ -63,6 +73,32 @@ export default {
...
+    /**
+     * Resize the sidebar, if the 'resizable' option is enabled.
+     */
+    onResize () {
+      const me = this;
+      const sidebarEl = me.$refs.sidebar.$el;
+      sidebarEl.style.transition = 'initial';
+
+      // Resize on mouse move
+      function onMouseMove (e) {
```
JakobMiksch, line 86: can we use a arrow function here instead?
then we do not need to have the `me` variable, because `this` can be used in the function directly

JakobMiksch (approved):
Works for me and code looks fine. 
One minor note regarding formatting and a question inline. Otherwise fine to merge.
Thx for the contribution and happy new year :-)

app-starter/components/AppSidebar.vue
fschmenger: To my knowledge declaring an arrow function and inlining it to the event handler only works, if the event doesnt have to be unregistered. So I could do it for `onMouseUp ` but not for `onMouseMove`. For clarity I decided to declare both handlers explicitly. 
I can give it a try though.

fschmenger:
Thanks for the review and a happy new year Jakob!

app-starter/components/AppSidebar.vue
fschmenger: Ok my bad. Obviously it works as well with arrow functions. You just can't inline them. Want to give it a quick test/ approval again before I merge?

JakobMiksch (approved):
@fschmenger the change in https://github.com/wegue-oss/wegue/pull/313/commits/a9d7a5c3b4b73af330e134be38e9cb73dba2b2eb looks fine to me. Please merge at will.

PR #314, opened 2023-01-25
Update to last Vue CLI v5 template

sronveaux:
Hi team,

As I had a little time, I took it to create this last PR in the `Vue-CLI` upgrade series.

This PR is again way smaller than https://github.com/wegue-oss/wegue/pull/312 and should be way shorter to review too... the most difficult part was to make `karma-webpack` work.

This one is based on the latest Vue CLI V5 template (5.0.8) which was published on the 7th of July 2022.

To stay in tune with https://github.com/wegue-oss/wegue/pull/297#discussion_r962979850, minimal versions of node and npm have been set according to the [official Vue migration guide](https://github.com/vuejs/vue-cli/blob/dev/docs/migrations/migrate-from-v4.md) where it is stated that support for `Node.js` versions 8-11 and 13 were dropped. So once again I took

    "engines": {
        "node": ">= 14.15.0",
        "npm": ">= 6.14.8"
    }

to use the first LTS version among the V14 ones.

Cheers
Sébastien


fschmenger:
Hi @sronveaux. 
Thanks a lot for providing this so quickly. I`m currently a bit limited on time, but hopefully can have look into it in the upcoming weeks.

sronveaux:
Hi @fschmenger,

Thanks for your reply, take your time, there's no hurry here. I had a little time and everything was ready (just a few commits to cherry-pick) so it's here.

Hope you'll like it !

Cheers,

fschmenger:
Hi @sronveaux. 
I finally went through the changes step by step and tested everything in a larger custom application First of all thanks alot for the fabolous work! I can see that the series of migrating Wegue to a stable build environment is finally coming to an end. Very nice also, that the previously introduced workaround related to Open SSL MD4 support is no longer necessary. So this runs stable under Node 18 LTS.

Regarding Karma and watch mode: The unit tests work fine in headless single-run mode. @chrismayer Could you give us some insights on what is supposed to work in watch mode, since I have no experience with that. See also `Stepwise Debugging chapter of the readme`.

I will add some comments inline, which are mostly questions. Maybe you have a second look there. 

Here is a summarization on the topics we discussed thoughout the previous PRs. I opened up some issues so we don't loose track of those. If I forgot something feel free to remind me :)

* #322: We could incoporate this in a follow up PR. If you have time to work on it, feel free to do so :) 
* #323: 
* #324

jsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "module": "esnext",
+    "baseUrl": "./",
+    "moduleResolution": "node",
+    "paths": {
+      "@/*": ["src/*"],
+      "APP/*": ["app/*"]
```
fschmenger, line 9: Could you give me some insights why the alias paths go in here, since this appears to be partially redundant to what we're doing in `vue.config.js::chainWebpack`?  I noticed this code is part of the cli template, so theres nothing wrong with it. I'm curious because I tried to remove the paths here and it if I recall correctly it didn`t have any effect. Same vice versa. My custom app introduces some other aliases, which I had to add in chainWebpack. Declaring them here did nothing for me.

package.json
```diff
@@ -11,104 +11,64 @@
...
     "ol": "6.4.3",
-    "proj4": "2.5.0",
-    "vue": "2.6.11",
+    "proj4": "2.8.1",
+    "vue": "2.6.14",
     "vue-i18n": "^8.24.4",
     "vuetify": "2.2.18"
   },
   "devDependencies": {
-    "@babel/core": "^7.0.0",
```
fschmenger, line 34: When I created a vue-cli template app based on v5.0.8 it used the following dev-dependencies. Did I accidently not use the most recent version of vue-cli or did you tweak dependencies manually to newer versions?

```
"devDependencies": {
    "@babel/core": "^7.12.16",
    "@babel/eslint-parser": "^7.12.16",
    "@vue/cli-plugin-babel": "~5.0.0",
    "@vue/cli-plugin-eslint": "~5.0.0",
    "@vue/cli-service": "~5.0.0",
    "@vue/eslint-config-standard": "^6.1.0",
    "eslint": "^7.32.0",
    "eslint-plugin-import": "^2.25.3",
    "eslint-plugin-node": "^11.1.0",
    "eslint-plugin-promise": "^5.1.0",
    "eslint-plugin-vue": "^8.0.3",
    "vue-template-compiler": "^2.6.14"
 }
```

vue.config.js
```diff
@@ -0,0 +1,60 @@
+const path = require('path')
```
fschmenger, line 1: The structure I got from template v5.0.8 looks a bit different.
```
const { defineConfig } = require('@vue/cli-service')
module.exports = defineConfig({
  transpileDependencies: true,
  lintOnSave: false
})
```

First of all it does not assign the configuration object to `module.exports` directly but invokes `defineConfig` on it.
Secondly it sets `transpileDependencies: true`, which I cannot see in here.

chrismayer:
>> Could you give us some insights on what is supposed to work in watch mode, since I have no experience with that

No, I have no experience with watch mode / stewise debug for the unit tests either. For me it was enough to call the test suite most of the times I developed tests. The https://github.com/wegue-oss/wegue/blob/master/test/README.md  was introduced by @justb4.  Maybe he knows something about it. If it does not work properly with our tools anymore it might be OK to remove and make an issue of re-adding this debugging feature for v2 if someone really needs it.

fschmenger:
Hi @sronveaux

I had a closer look at the unit tests again, and currently I`m getting this warning:

```
karma-webpack does not currently support custom entries, if this is something you need,
consider opening an issue.
ignoring attempt to set the entry option...

karma-webpack does not currently support customized filenames via
webpack output.filename, if this is something you need consider opening an issue.
defaulting static/js/[name].js to [name].js.
```

There is also a more severe bug when moving to OL 6.15.7- see #320.

sronveaux:
Hi @fschmenger, hi @chrismayer 

Thanks for the review and happy to see it is appreciated !

Concerning the `Karma` watch mode, as explained once, I'm pretty sure all works as expected, you can manage to break it but it comes from an incorrect use of `Karma` itself. What needs to be done in order to break it in this PR also breaks in current `Wegue` version, so this is not a regression introduced by this PR...
We just have to keep in mind that running in this mode is an intentional command typed by the user (or an intentional change in the `karma.conf.js`), so the only thing that can be done here is remove the explanation in the documentation, but nothing hard-coded in `Wegue` should be problematic for this I think.

@fschmenger concerning the warnings you see in Karma-Webpack, these should be irrelevant in our case as I'll explain just down here. The potential issues they warn about are very specific and would only happen in complex scenarii as you'll certainly understand. Also, given the fact that the latest version of `karma-webpack` is from February 2021 and nobody requested those addition to the author as stated in the warnings, it really looks like nobody is impacted. Some issues are still opened on their Github repo and discussions are followed so this plugin is not dead so if someone had problems with that, we'd certainly find something but there's nothing.
Why those warnings shouldn't introduce bugs is explain down here:

1. Concerning `karma-webpack does not currently support customized filenames`: This warning is a nice addition in the new version of `karma-webpack V5` which in fact tells us what it has always done behind the scenes silently. If you look at source code of `karma-webpack V4`, you'll find [this](https://github.com/ryanclark/karma-webpack/blob/48878ebc119df35d91da2f9ecf4c8b797ecc4a45/src/karma-webpack.js#L104:L105). Just that now there is an `if` warning us that the `output.filename` won't be taken into account during testing as it was always the case before.

1. Concerning the `karma-webpack does not currently support custom entries`: Here also, this warning is a kind addition in the new version of `karma-webpack V5`. It comes from the fact that we have `entries` automatically defined for us in the Webpack config by `Vue` which is `almost` normal as we are in [this scenario](https://webpack.js.org/concepts/entry-points/#multi-page-application). (I said almost as our different pages are linked to the same entry point in our case so we should ideally have only one entry, but that's how `Vue` builds the `Webpack` config while reading the `vue.config.js`... that's not our point here.)
However, for `Karma` to run properly, it has to build its custom package where the entry point is the test runner and not `Wegue` itself. You can get a general idea of what's happening [here](https://stackoverflow.com/questions/36766624/how-to-use-webpack-multiple-entries-with-karma). The idea is that all what is required from `tests/unit/index.js` and its dependencies (each and every specfile in our case) will be bundled without worrying about what the entry point of the project is.

We could potentially remove those warnings be changing the webpack config manually when it's loaded in `tests/karma.conf.js`. However, this would be doing what the warning tells us `karma-webpack` is doing and would, I think, hide the fact that the `Webpack` config given to `Karma` is not the one we have written at first. As they are simple additions around how the plugin always worked, I don't think if worths the pain given the fact that if someone is doing very specific things inside it's app, hiding those warnings would hide potential investigation paths to solve their problems.

Unfortunately, I'm almost 100% sure the #320 error is not linked to those warnings and should be addressed apart... I'll comment on this issue as you requested as soon as I'll have time...


At last, concerning the conversations you opened, I'll reply separately as soon as I'll have time to check before telling silly things ;-)

Thanks again and hope this quite lengthy explanation will be useful to all,

Cheers,
Sébastien

jsconfig.json
sronveaux: Funny that your conversations are in order of difficulty to answer here ;-) this one is not too hard to explain... well perhaps...

`jsconfig.json` is a file which at a time was specific to `VSCode` and was here to give information to its `LSP` about the project. You can find the [official documentation on VSCode here](https://code.visualstudio.com/docs/languages/jsconfig) and [the PR and discussions why it was added in Vue-CLI here](https://github.com/vuejs/vue-cli/pull/6285).
It's said there that the file was added for [Vue-CLI to work better with Vetur or VueDX](https://github.com/vuejs/vue-cli/issues/6064#issuecomment-733543301)
This file is since used by other code editors too like `SublimeText` and is here mainly to give better coding experience and provide, for example, auto-completion while writing imports.
That's why nothing happens when you modify this file, it doesn't impact the build, only the way your code editor behaves... if it makes use of such a file.

I personally use it in some projects to type hint my code with `JSDoc` and have the code editor warns me when the types are not respected (see `checkJs` in the doc) which would be very useful in `Wegue` but would take so much time to implement...

Eventually, there are commonly two things to be patched inside this file, first the alias paths for the reason given up here (autocompletion inside the code editor) and the exclude key so the code editor don't think `node_modules` is part of our project and analyses it which can take ages...
Here's [a small article](https://lmichelin.fr/vscode-intellisense-jsconfig-vue-js-nuxt-js-absolute-imports/) which says this better than I am for completeness.

package.json
sronveaux: No I think we are synced here, just look 6 lines down the `package.json` diff, the line you commented is a removal of `"@babel/core": "^7.0.0"` and 6 liner later, `"@babel/core": "^7.12.16"` is added as in your template.

This comes from the fact that I didn't generate an empty project from scratch and copy `Wegue` parts inside but started from `Vue-CLI v4 PR` and upgraded `Vue` components as stated in the [official migration guide](https://cli.vuejs.org/migrations/migrate-from-v4.html).

There are some discrepancies for sure, the 5.0.8 template for example still installs `"@vue/cli-plugin-babel": "~5.0.0"` as stated in your comment where the upgrade turned this automatically into `"@vue/cli-plugin-babel": "~5.0.8"`. However, those discrepancies should at least stick with what the semver syntax says. I mean ^7.0.0 and ^7.12.16 can potentially be the same as ^7.x.x says that any package < 8.x.x can be installed if required to by a dependency and ~5.0.x means < 5.1.x.

vue.config.js
sronveaux: O.K., this one is the hardest to explain and I had to do some tests to remember where this all comes from.

Concerning the call through `defineConfig` first, this is something new added in `Vue-CLI v5` which is kind of optional as stated in [the official documentation](https://cli.vuejs.org/config/#vue-config-js).
You can see that we can use two different formats and that I used the first one instead of the second. In the source code, we can see that [defineConfig](https://github.com/vuejs/vue-cli/blob/v5.0.8/packages/%40vue/cli-service/lib/Service.js#L375:L376) doesn't do anything else than type hint the content of the object inside, which is also done explicitly in the first version.
The official documentation states that `defineConfig could provide better intellisense support`. However, if you look [in the source once again](https://github.com/vuejs/vue-cli/blob/v5.0.8/packages/%40vue/cli-service/types/index.d.ts#L136:L138), this would only be relevant if a function is used instead of an object which is not our case.
In the rest of the documentation, the good old `module.exports` is used in all examples. So at that time, I decided to stick to the first format which is IMHO much clear and transparent compared to calling a function burried deep inside the code and which is not well documented.
Moreover, as this `defineConfig` isn't used anywhere else in the official documentation itself, I personally think it makes you wonder why the explanations never use it and why you have to write your config file differently than what the official documentation explains. However, this can easily be changed if you want it though...

`transpileDependencies: true` is linked to the same documentation discrepancy thing but has a more deep impact that we can argue here:
This parameter is linked to which modules from `node_modules` should be transpiled through `babel-loader` and which shouldn't and should be kept as is. I think the two major which concerns us here is `Vuetify` and `OpenLayers`.
First, the template sets it to true as you mentioned up here but the [official documentation](https://cli.vuejs.org/config/#transpiledependencies) says this defaults to `false` if not specified and that `transpiling all the dependencies could slow down the build process` and let you specify an array of dependencies to transpile.
When I first made the changes for this PR in first half of 2022, I remember that dependencies installed at that time made errors while transpiling `Vuetify` so I had to remove this line. Today it's not the case anymore, only warnings can be seen, so it's true that we could potentially put this line back to stick to the official template... which diverges from the official documentation.

When generating a new `Vuetify` app instead of a pure `Vue` one, this line is changed from `transpileDependencies: true` to `transpileDependencies: ['vuetify']`. So I thought about looking at this when including the `Vuetify` plugins and `ESLint` support...

However there's another point to think about beforehand today:
Before Vue-CLI v5, when building your project, it was passed through `Babel` normally, but you could also have your app built in [modern mode](https://cli.vuejs.org/guide/browser-compatibility.html#modern-mode). There, `Vue CLI` would produce two versions of the app, one legacy bundle transpiled to `ES5` and one modern bundle aiming `ES2015+` code, both added in the generated `html` files. With a simple technique, modern browsers are able to skip the legacy bundle and load only the modern one.
`Vue-CLI v5`, [has made the modern mode the default one](https://cli.vuejs.org/migrations/migrate-from-v4.html#the-build-command-and-modern-mode). And that's where funny things happen, when `npm run build` is launched today, with its new way of working which takes into account the content of `browserlist`, we can see those two lines appear:

> All browser targets in the browserslist configuration have supported ES module.
> Therefore we don't build two separate bundles for differential loading.

It means the generated app is now only available to `ES2015+` compatible browsers. We can still transpile some dependencies so that potential polyfills for `ES2022` or `ES2023` novelties are added, but this won't have the same impact as before where all was transpiled to `ES5`.
Note also that we can force the generation of a legacy build by passing a `--legacy` flag to the build command.

[`Vuetify` official documentation](https://v2.vuetifyjs.com/en/getting-started/browser-support/#browsers) states that only `IE11` and `Safari 9` would require the use of polyfills, both of which are not taken into account with current `browserlist` file. So it means adding `Vuetify` in `transpileDependencies` changes nothing in terms of compatibility with current configuration, except we can see some warnings telling us transpiling makes worse than better:

> [BABEL] Note: The code generator has deoptimised the styling of C:\Sources\Wegue\UpdateCLI\node_modules\vuetify\dist\vuetify.js as it exceeds the max of 500KB.
> [BABEL] Note: The code generator has deoptimised the styling of C:\Sources\Wegue\UpdateCLI\node_modules\@mapbox\mapbox-gl-style-spec\dist\index.es.js as it exceeds the max of 500KB.

I haven't looked for `OpenLayers` though, knowing that the current version has certainly not the same needs as the one we're trying to update to in #320 ...

So to summarize, we can argue here with such questions:

- Should we modify the `browserlist` file to target a broader range of browsers and more polyfills are added while bundling or should we stick with the current content ?
- Should we prefer to set `transpileDependencies: true` which has no effect in terms of compatibility with current `browserlist` configuration (except perhaps for `OpenLayers`, to be checked depending on the version we're aiming for) but warns that styles are not optimized anymore or should we aim for maximum compatibility ?
- Should we aim only modern browsers with the release of `Wegue V2` and so remove this `transpileDependencies` by default, setting a note in the documentation about this and explaining to users what should be done in case of compatibility issues ?

I don't know what you and @chrismayer think about this, that's only a single line to change in the end in terms of code, but gives a very different message to the people who will use `Wegue V2`...

Sorry for the waaaaay tooooo long comment once again and hope this is all understandable...

Cheers,
Sébastien

jsconfig.json
fschmenger: OK, thanks a lot for the explanation. And sorry in hindsight for not doing investigations on this myself :)

package.json
fschmenger: OK, I just wanted to make sure that this hasn`t been an oversight. The dependencies all work out and if I got you right, this is what `vue upgrade` has done to it. All good here!

sronveaux: Exactly, 100% correct!

vue.config.js
fschmenger: Thanks a lot Sebastien for the very conclusive explanation.

I myself tumbled over the `browserlist` build target console print, which told me there's currently only output for the 'ES2015+' target. It`s a good question, whether we should support legacy browsers. My personal opinion is, since this hasn't been supported previously we could just leave it as is. If it is required then we could address that later on in a separate PR.

As you proposed, I would just stick with the structure of the template here.
* Invoke `defineConfig` - even though it's a no-op.
* Then set `transpileDependencies: false` explicitly , maybe with a short comment for the reasoning as pointed out above and that it could be granularly enabled. I totally agree with you, that we should not impose a compile overhead, when there is absolutely no need to.

This way it's immediately apparent where we are going down our own track, just in case we have to pinpoint that later.
Note: There is no problem related to OL when not transpiling dependencies, more later.

fschmenger:
Hi @sronveaux, thanks once again for the verbose explanation on how the build system behind Karma-Webpack works!

Regarding the warnings: Those are no biggie. I assume that the entrypoint of the application will never be required for the unit tests. So we could just get rid of the warnings as you proposed in your snippet in #320. 

Regarding the build error with the new OL (described in #320):
With the help of your hints, I could finally track it down to Webpack Config using the wrong build `target` for the unit tests. It has been set to `node` where I it should be set to `web` to produce a bundle that can run under Chrome Headeless. This is the difference to the production build, which uses `browserlist` as a target and caused all the issues. See [webpack targets](https://webpack.js.org/configuration/target/)
So I fixed it with the following snippet in `karma.conf.js`:
```
const webpackConfig = require('../../node_modules/@vue/cli-service/webpack.config.js');
webpackConfig.target = 'web';
```

To give a brief explanation why it is hitting us on the OL migration:
OL now includes a module called ['web-worker'](https://www.npmjs.com/package/web-worker). This comes in 2 versions, 1 for node `/cjs/node.js` and another for browser environment `/cjs/brower.js`, where in the latter case it's simply an alias to `Worker`. This is probably the first time that the wrong build target mattered, which is apparent from the following warning:
```
WARNING in ./node_modules/web-worker/cjs/node.js 214:8-20
Critical dependency: the request of a dependency is an expression
 @ ./node_modules/geotiff/dist-module/worker/decoder.js 2:8-40 5:21-27
 @ ./node_modules/geotiff/dist-module/pool.js 49:8-37
 @ ./node_modules/geotiff/dist-module/geotiff.js 5:0-29 732:0-16
 @ ./node_modules/ol/source/GeoTIFF.js 21:0-108 77:25-29 189:18-30 192:18-30 195:18-29
 @ ./node_modules/ol/source.js 10:0-57 10:0-57
 @ ./node_modules/babel-loader/lib/index.js??clonedRuleSet-40.use[0]!./node_modules/@vue/cli-service/node_modules/@vue/vue-loader-v15/lib/index.js??vue-loader-options!./src/components/ol/Map.vue?vue&type=script&lang=js& 22:14-34
 @ ./src/components/ol/Map.vue?vue&type=script&lang=js& 1:0-233 1:249-252 1:254-484 1:254-484
 @ ./src/components/ol/Map.vue 2:0-55 3:0-50 3:0-50 14:2-8
 @ ./src/ sync ^\.\/(?%21main(\.js)?$) ./components/ol/Map.vue ./components/ol/Map
 @ ./tests/unit/index.js 33:19-78
```

Note: I tried to fix it in `vue.config.js chainWebpack`, but if I do
```
if (process.env.NODE_ENV === 'test') {
   config.target = 'web'
}
```
then it somehow gets rolled back under the hood to `node`.

All of this results in the following questions:
* Can we somehow consolidate what's in `vue.config.js chainWebpack` and the (potential) modifications to the config from `karma.conf.js` ? I think, if possible the webpack configuration for testing should not be scattered over 2 places. So we should either do the fixes mentioned above in `chainWebpack` (if it can be done), or move everything over to `karma.conf.js`.
* As you pointed out this is probably not a regression in this PR. Do you mind fixing it in here, since it`s somehow in the scope of the build ecosystem, or should we do it separately?

Greets,
Felix

fschmenger:
I forgot one thing connected to the same topic: You pointed me to [the karma-webpack source](https://github.com/ryanclark/karma-webpack/blob/48878ebc119df35d91da2f9ecf4c8b797ecc4a45/src/karma-webpack.js#L113). As far as I can see it already should turn `splitChunks` and `runtimeChunk` off. You have any idea why it doesn't work when I remove the corresponding lines in `chainWebpack`?

sronveaux:
Hi @fschmenger,

Glad that my boring explanations were not useless and helped you find the problem we had here !

For sure, I can try to solve this here as this is the most logical place to do so. I just have very few time for now and will do it as soon as I can. I'll have to delve a bit more inside this and see how we could bring all together cleanly.

This totally makes sense and is unfortunately a big problem with what they've done with `Vue-CLI`. You don't write the `Webpack` config directly but have to define things in `vue.config.js` following their format, then if you have to do more advanced things you have to use `chainWebpack` but you don't have access to the whole config file directly as it is generated on-the-fly for you. [The official documentation](https://cli.vuejs.org/guide/webpack.html#using-resolved-config-as-a-file) says that you require a certain file that `dynamically resolves and exports the exact same webpack config used in vue-cli-service commands` except it is not in all cases.
I think this comes from the fact that we are in `test mode`, instead of `production mode` or `development mode` and the fact that the recommended test frameworks are now `Mochapack` and `Jest` which both are node-oriented and use `JSDom` instead of a headless browser by default.
I'll make some tests and try to figure out the best way to make this work then come back to you...

Concerning your question on `karma-webpack`, the link I pointed you to and that you referenced in your question is a link to the `karma-webpack V4` codebase which was used in `Vue-CLI V4` and is compatible with `Webpack4` but not with `Webpack5`. That's why you have to update it to `V5` when you use `Vue-CLI V5`.
`V5` is a complete rewrite of the plugin and the two options you are mentioning are not removed anymore as in V4. They are even [set by default](https://github.com/ryanclark/karma-webpack/blob/v5.0.0/lib/webpack/defaults.js#L17:L30) in this version.
I can't remember exactly why this was a problem but I clearly remember struggling with it. And I'm not the only one, you can still find some issues in their repository like [this one](https://github.com/ryanclark/karma-webpack/issues/506) or [this one](https://github.com/ryanclark/karma-webpack/issues/491).
I remember I came across a repository of a modified version having only those lines changed. But I didn't want to use a patched version which was not published and would not be maintained afterwards...
It was linked to the official explanation [here](https://github.com/ryanclark/karma-webpack#how-it-works) and the fact that since `V5` they decided to generate multiple builds to optimize things. But this proves to be more problematic than positive unfortunately...

Cheers,
Sébastien

sronveaux:
Hi @fschmenger,

Good news, I had a very short time and managed to correct what you found directly inside the `vue.config.js`, I managed to make the unit tests run on an updated version of OpenLayers!
I just will have to find the best and cleanest way to do it by delving a little bit more inside `Vue-CLI` code but this can be be done!
I'll keep you updated as soon as I have time to look more precisely and re-test the whole thing!

fschmenger:
Hi @sronveaux. Excellent that you are looking into this! There`s no urgency at all so take your time.

Cheers,
Felix

sronveaux:
Hi @fschmenger,

I had a little time and managed to go to the bottom of it easier than expected. It is indeed directly linked to what I thought at first, the fact that the two recommended test frameworks are node-based and use `JSDom` by default. In the code, you can even find some clues like [this](https://github.com/vuejs/vue-cli/blob/aad72cfa7880a0e327be06b3b9c3ac3d3b3c9abc/packages/%40vue/babel-preset-app/index.js#L107:L110).
When you use `Vue-CLI` either directly (by calling `npm run serve` or `npm run build`) or indirectly by requiring the webpack config file, a lot of things happen under the hood in a way not really documented...
To summarise, [the `Webpack` config is built in two steps](https://github.com/vuejs/vue-cli/blob/8b5ab227627a3bfbea31b24364f060154a9b0b89/packages/%40vue/cli-service/lib/Service.js#L265:L289), first you have the `webpack-chain` machinery used to create a `raw config`, which is mainly defined under the [chainWebpack key inside vue.config.js](https://cli.vuejs.org/guide/webpack.html#chaining-advanced). This `raw config` is then amended with what you define under the [configureWebpack key inside vue.config.js](https://cli.vuejs.org/guide/webpack.html#simple-configuration).
These two steps are built from an array of callbacks which are registered from what we define in the `vue.config.js` file, but can also be registered by any of the installed `Vue` plugins. And that's where the [Mocha plugin registers a callback that uses chainWebpack to add the target="node" key](https://github.com/vuejs/vue-cli/blob/70afd90ca357ee93823c61ccb1378d5f00eb063a/packages/%40vue/cli-plugin-unit-mocha/index.js#L2:L9).
As this is done in the `webpack-chain` step, we can then apply some changes to it from the `configureWebpack` step afterwards...
I thus propose to do these test config amendments from here, remove the `target` key as is the case in other modes (`web` being the default one in `Webpack` it is not added in `development` and `production` modes neither), and potentially the two ones needed to remove the warnings given by `karma-webpack` if you think it is better to do so. Does it make sense for you?

Cheers,
Sébastien

sronveaux:
Hi @fschmenger,

Thanks for the comments,

I just took a look to the remaining errors in unit tests and can assure you it has nothing to do with the generated bundle.
It is linked to the formula implemented in `OpenLayers` used to generate the `SCALE` parameter if not specified in layerSource construction. Currently used `V6.4.3` has [this one](https://github.com/openlayers/openlayers/blob/v6.4.3/src/ol/source/TileWMS.js#L262:L269) and since `V6.5.0`, [this one is used](https://github.com/openlayers/openlayers/blob/v6.5.0/src/ol/source/TileWMS.js#L262:L268) which gives different results in lower decimals.
So this is an `OpenLayers` upgrade bug, not a `Vue-CLI` one and should be corrected somewhere else, like in #320.

I will certainly take the time to amend this PR next week with all the discussed contents, however I had a simple question/suggestion beforehand:

In what you've written in your comments up here, I would replace the `config.target = 'web'` line with `delete config.target` as the other two generated configs (development and production) don't assign a value to it. This would generate the closest possible configurations for the three modes.
If you think explicit is better than implicit, I would suggest to move the line outside the `if` clause and set the value for allthree configs then... what do you prefer?

Have a nice day,
Sébastien

sronveaux:
Hi @fschmenger,

Thanks for the fast reply!
100% agree with you, I'd also prefer to tweak the `test` config only and certainly, if `Wegue` stays with `Karma` (which I personnally think is a good choice), this target will need to be `web` whatever happens in the future.
Concerning the potential removal of the line or the explicit set, I'll ask the question differently to take the decision then, do you think `Wegue` would ever need to use anything else than a `web` bundle? If you think it can, then I'd set the line, if not, I'd remove it. Sure it can break if `Webpack` team changes something under the hood later, but this change would also break the build itself and would then need this setting to be applied globally...

However, I'm pretty confident this won't change until `Webpack 6` which will bring a lot of breaking changes (as always with it) and will eventually need some tweaks here and there... or perhaps will `Wegue` transit to `Vite` before that and use `Rollup` behind the scenes... way too soon to discuss though, that's another history...

fschmenger:
Hi @sronveaux,
I think there are no short term plans to exchange Karma, as there would be a lot of work involved.

sronveaux:
Hi @fschmenger,

Found the time before a meeting to amend the PR with all the comments.
This should be it, it will just need to be amended and merged with new additions that would come before being accepted!

Cheers,
Sébastien

fschmenger:
Hi @sronveaux. Thanks for finalizing this! Looks like you forgot 1 important thing :) See my comments inline.

Cheers,
Felix

sronveaux:
Hi @fschmenger,

Stupid boy I am, forgot THE important line here!  I had it so I suppose something went wrong during the cut / paste operation!
The funniest thing is that I re-tested the whole thing before pushing, but as `OpenLayers` is still `V6.4.3`, no problem occured!

I took the opportunity to tune one thing that came through my mind a few days ago and synced following the merge of #325 so this one should now be ready to go and ready to merge in upcoming V2!

fschmenger:
OK great, looks like we`re finally there. 
Thanks @sronveaux for all the work that has gone into this!

sronveaux:
And thanks for your confidence, patience and time reviewing all those cumulative PRs!

chrismayer (approved):
Thanks a lot to all, who were involved here.

sronveaux:
Hi @chrismayer,

So good to see that `Wegue V2` will become a reality soon! Thanks for the job!

Just a little warning here, I suppose you only merged PR #320 this morning... my PRs #298, #297 and #312 are marked as `merged` (because they were inside of #320) however my latest and more up-to-date PR #314 (which I'm replying in) is still marked as `open`. I suppose this is because some of the commits were not merged inside #320 but cherry-picked.
However, we should take great care and double-verify that all commits of #314 have been taken into account by @spwoodcock in #320 to avoid the latest troubles we've hunted down with @fschmenger...
If all is present inside current `Wegue` master branch, then we can close this PR...

Thanks again and wish you a pleasant day,

Cheers
Sébastien

spwoodcock:
You are right @sronveaux, I had merged in up to fc2936a, and then cherry picked f9b0828.

But I'm not sure if the commits after that have been included, my apologies. 

One option could be to rebase this branch with master to see which commits are missing after.
Or we could just cherry pick in the commits after f9b0828.

chrismayer:
Thanks for your important feedback @sronveaux and @spwoodcock. I tottally agree that we should carefully check everything is in.

@sronveaux could you please rebase this PR with current master as suggested by @spwoodcock ? Would be really cool and we would easily see, if there is something missing.

fschmenger:
The problem is, if you rebase everything and push-overwrite, you might lose a lot of information in here like comments to specific check- ins. So I suggest you either setup a new PR which contains everything based on master. The simpler approach could be to just merge the master with this PR again and make it available as a new PR. I unfortunately have no time to study the details at the moment.
Cheers Felix

chrismayer:
Thanks @fschmenger for sharing your concerns. So just merging in the current master into the base branch of this PR should be fine and less invasive.

@sronveaux Could you do that? If this is not possible from your side I could do this but I would have to do in a new branch since I have no write permissions for your branch.

sronveaux:
Hi @chrismayer @spwoodcock @fschmenger,

Thanks for your returns on this.

@spwoodcock don't worry, things like this happen, it shouldn't be too hard to fix in the end!
@fschmenger thanks for your comment, I would have avoided to rebase the whole thing too and am happy you confirmed my position,
@chrismayer, I'll take a look at this, either this afternoon or on Friday and keep you updated as soon as its done!

Cheers
Sébastien

sronveaux:
Hi @chrismayer,

I just merged the current master as expected. I also verified the following:

- No lint errors are present
- The project builds correctly
- All unit tests pass
- Docker image can be built and run flawlessly

That should do it. I just took the opportunity to correct the `package-lock.json` file which was turned into the V2 format after this PR was released but was still tagged as V1. `Wegue` was also listed as still being V1.2.  
You can easily review it, it's a 3-line change.

Hope this is now ready to go!

Cheers,
Sébastien

chrismayer:
I just tested this with a recent application based on the current master. Everything looks good, the build works and the test suite passes. Therefore I'll happily merge this.

Thanks again to all who were involved and especially @sronveaux for this last extra step to make this mergeable again.

PR #316, opened 2023-02-02
Configure "X" button to close module windows

chrismayer:
This introduces the module config property `closable` in order to configure whether the module window can be closed by a "X" icon in the window's header bar or not. The default is `true`, so previous behavior (always render the X icon) is not changed for existing configurations.

### DIFF
--- a/app-starter/static/app-conf-sidebar.json
+++ b/app-starter/static/app-conf-sidebar.json
@@ -223,7 +223,8 @@
       "win": "sidebar",
       "icon": "layers",
       "visible": true,
-      "minimizable": true
+      "minimizable": true,
+      "closable": false
     },
     "wgu-measuretool": {
       "target": "menu",
--- a/app-starter/static/app-conf.json
+++ b/app-starter/static/app-conf.json
@@ -181,7 +181,7 @@
       "format": "MVT",
       "attribution": "Kindly provided by @ahocevar",
       "visible": false,
-      "opacityControl": true,  
+      "opacityControl": true,
       "style": {
         "strokeColor": "gray",
         "strokeWidth": 1,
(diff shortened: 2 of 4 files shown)


fschmenger:
Hi Christian,
first of all I wonder if the terminology `closable` could be a bit irritating to the user, as obviously we can still toggle the module visibility via the toolbar. On the other hand naming is in sync with the  `minimizable` property (keep in mind that here the functionality is only available via the module card header though). 
I was thinking about providing an option to hide the window header entirely, which could be useful and maybe less confusing.

So I`m a bit uncertain whether this is a useful feature. But if you have a requirement to hide the close button, then go ahead with it.

Codewise it`s fine. Could you
* add 1 line in the ModuleCard unit test `has correct default props` block
* maybe sync the example config at the end of `wegue-configuration.md` with the modified` app-conf.json`

Anyway works as designed so I approve it.

fschmenger (approved):

chrismayer:
Hi @fschmenger ,

thanks for your feedback. Indeed I had 2 or 3 times a customer request to hide the X button in the window header. I guess the point is to have a single point (the button in the toolbar) to disable / enable the module, which is implicitly done when a user clicks the X button. As I understood this confuses users since the X does not only hide/close the window. One surely could hide the X button via CSS on application level, but since I had this several time I thought bringing this into Wegue would be more sustainable. The name was actually based on the `minimizable` and inspired by some other UI libs. If you have a more appropriate name I happily would take this. Something like `closableByWinHeader` seemed a bit unhandy for me.

Since this a tiny optional feature and if no one sends "-1" I vote to bring this in for the mentioned reasons.

JakobMiksch (approved):
The change works well and can be merged from my side.

fschmenger:
Hi Christian, thanks for the explanation. +1 to merge then.
Cheers Felix

chrismayer:
Thanks for your feedback @fschmenger and @JakobMiksch...I going to merge now.

PR #317, opened 2023-02-03
Extend 'sidebar-toggle' event to force open state

chrismayer:
This extends the `'sidebar-toggle' `event of the `AppSidebar` component, so the opening state of the sidebar can be forced (not just toggled). Done by passing a boolean argument when firing the event. Harmonizes the signature / behavior with the ` 'app-loading-mask-toggle'` event, which does similar things.

### DIFF
--- a/app-starter/components/AppSidebar.vue
+++ b/app-starter/components/AppSidebar.vue
@@ -16,16 +16,16 @@
           absolute
           top
           color="secondary"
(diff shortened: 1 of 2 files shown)


JakobMiksch (approved):
Useful addition! 
Since, it does not change the default behavior, I vote for merge.

chrismayer:
Thanks for the review @JakobMiksch. I am going to merge...

PR #319, opened 2023-02-16
Upgrade MDI libraries

chrismayer:
This upgrade the Material Design Icon libraries used within Wegue to their latest versions in order to get a set of new icons:

  - @mdi/font => ~~7.1.96~~ 7.2.96
  - material-icons => ~~1.13.1~~ 1.13.9

No functional changes, also tested in build. See screenshots below:

_Before:_

image

_After:_

image

/cc @spwoodcock @sronveaux @ntma


fschmenger:
Hi Christian, thanks for providing this.

I already did it for my custom application and had a similar commit prepared to transfer it to Wegue. The problem here is, that the mdi-fonts package has a new major version and is no longer compatible with version 5.x, as they removed a couple of icons or changed namings. You can see that when diffing the related packages in your node_module folder.  It might not break Wegue but will break Wegue based applications which have those old icons in use.

So I suggest we put this on hold, and incorporate it in Wegue v2.

Greets Felix

chrismayer:
Thanks for sharing your experience on this @fschmenger. I did not discover any problems at Wegue levels and the apps I am currently maintaining, which is surely only a limited view. So sorry for the noise.

fschmenger:
Hey Christian, no noise involved at all.

We probably need some pattern for managing all the PRs that will go into Wegue v2, so we don't lose track. Any ideas how we should accomplish that?
E.g. we could set up a repo, which pools all the v2 PRs. I`m not sure whats the best practice here and currently limited on time to do some investigations. If you have any ideas, feel free to share. For the time being you could just leave this PR open, so it will not be lost in space.

chrismayer:
> We probably need some pattern for managing all the PRs that will go into Wegue v2, so we don't lose track.

Maybe we could introduce a label `v2` or similar and/or we could make a github board or a project (how it seems to be named now) to collect the issues. github projects are described as 

`Built like a spreadsheet, project tables give you a live canvas to filter, sort, and group issues and pull requests. Tailor them to your needs with custom fields and saved views.`

But maybe this PR is not the right spot for this discussion. Maybe a discussion (https://github.com/wegue-oss/wegue/discussions/) or the gitter chat would be more transparent.

fschmenger:
Hi Christian, as proposed I reopened it and labelled it v2.

chrismayer:
Reactivated this by rebasing to current master and setting latest versions of `@mdi/font` (`7.1.96`) `material-icons` (`1.13.1`).

chrismayer:
I did some additional tests and everything seems fine. I also added a fix for a missing icon within the `MapRecorder` module and some upgrade notes.

Please review.

sronveaux (approved):
Hi @chrismayer 

Reviewed all changes, gave it a try and updated some personal project with it to see how it fits and everything works perfectly !

The upgrade note contains all that should be needed, seems perfect on my side !

Cheers

chrismayer:
Thank you very much for the review @sronveaux. I am going to merge now.

PR #325, opened 2023-03-14
Maintenance v1.2.0

fschmenger:
This branch contains a couple of minor fixes. For details please see the commit messages.

### DIFF
--- a/app-starter/components/SampleModule.vue
+++ b/app-starter/components/SampleModule.vue
@@ -103,7 +103,7 @@ export default {
 <!-- Here we do the styling of our module -->
 <style scoped>
 /* our module has the class '.sample-module' and we reference it here */
-.sample-module {
+.sample-module.wgu-floating {
   left: auto !important;
   top: 300px !important;
   right: 10px;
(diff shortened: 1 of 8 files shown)


chrismayer:
Thanks for this @fschmenger. Code looks good and I tested the changes in a customer app and in the the example apps and one thing I discovered: The map recording module shows an error `Failed to start recording`.

image

I am unsure how this is related to the changeset, but it worked on current master which I tested before.

chrismayer:
I gave it another shot and it is related to the active layer set on the map. When activating the Tiled WMS layer in the demo apps the module won't work (even wont work anymore, if the layer is deactivated). Totally unrelated to this changeset.

So this is good to go. Feel free to merge. Thanks again!

chrismayer (approved):

fschmenger:
Hi Christian, 
it seems that both ahocevar layers are no longer CORS enabled. It used to work, as a quick search yielded 80f2167f2181f5eed980f17c60941eaaeae5901d . Nevertheless maintenance is a good anchor to address such problems, so I ammended the fix. 
Could you give it a quick a verification? Then feel free to merge.

chrismayer:
Thanks for the additional fix regarding the MapRecorder. I can confirm that it works now.

Please merge.

PR #320, opened 2023-02-28
Merge Vue CLI updates & upgrade to OpenLayers v6.15.7

spwoodcock:
**What is included:**
- First I have merged all Vue CLI PRs sequentially and resolved conflicts (#298, #297, #312, #314), great work @sronveaux! I hope this helps to save time in getting these great additions merged.
- Then I solved the remaining linting errors and upgraded to OpenLayers 6.7.0. I did some basic testing and couldn't find errors, so this could close #261.
- There are no breaking changes I can see in OpenLayers until v7, so I upgraded to v6.15.1, the last version prior.
- Upgrading to v7 would require some refactoring for removed methods: https://github.com/openlayers/openlayers/blob/main/changelog/upgrade-notes.md#700
- After merging this PR, implementing #290 would be possible.

**What I missed:**
- I couldn't get the Karma ChromeHeadless tests to work, as I run my coding environment inside a docker container.
I get (expected) errors related to system dbus and WebGL:
```
Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
Passthrough is not supported, GL is disabled
```

PR #333, opened 2023-04-26
Update Vue to v2.7.14 LTS + Vuetify v2.6.15

### LINKED ISSUE #330: Upgrade to Vue 2 and Vuetify 2 LTS
Wegue should be upgraded to Vue 2 and Vuetify 2 LTS.

spwoodcock:
This PR is built on top of #320, to allow for an easier single merge.
Apologies for this non-standard approach, but I couldn't think of a better way - these updates are reliant on the previous PR commits.

Closes #330.

- I upgraded Vue, as in the title.
- Also upgraded vue-i18n to the latest Vue2 version.
- Vuetify was upgraded to `v2.6.15`. IMO, we shouldn't wait for Vuetify `v2.7` LTS to be released, as Vuetify development can be painfully slow. It shouldn't really matter as `2.7` seems to be focused on "backports and deprecations from v3.0.0 to help developers prepare for v3 upgrade".

**Additional considerations**:
- I pushed a version of **package-lock.json** generated from `npm install` using NPM v6, as v7 updates the lock file to `lockfileVersion: 2`. Ideally all developers should be using the same version of NPM, but this is hard to enforce. I considered updating `engines.npm` in **package.json** to have a maximum version and creating a `.npmrc` with `engine-strict: true`, but this may cause issues with downstream users.
  - NPM v6 can be used by running `npm install` inside a `node:14` docker container.
  - NPM v6 is from 2018, so in future we could update to NPM v7-8 (`lockfileVersion: 2`), or NPM v8 (`lockfileVersion: 3`).
- I also updated the **vue.config.js** `devServer.port` to `0.0.0.0`, as it allows more flexibility than `localhost`.


fschmenger:
Hi Sam,
thanks for providing this. The next 2 weeks will be busy here, then I hopefully will find a time slot to have a closer look.

I don`t see any problem with PRs being built on top of each other, we used to follow the same approach with all the vue template upgrades. So, even if this incorporates #320, could you reopen #320 again? This will give us the opportunity to review and merge in changes in a more fine grained way. 

Cheers,
Felix

spwoodcock:
No problem, I will re-open 👍 
This PR is superseded by #334.

spwoodcock:
Merged `master` to only show specific commits - ready to test and merge.

spwoodcock:
Closed in favour of #335

Issue #290, opened 2022-04-13
cog geotiff support on wegue


pvgenuchten:
COG support was introduced on openlayers in v6.7
@JakobMiksch has prepared a PR with migration to v6.7 at #261, which is a prerequisite for this work
What is the status on this migration? 

I'll have a look at the build failures, to see if i can progress the work

JakobMiksch:
@pvgenuchten I started this PR to try COG functionality, but had problems with upgrading OpenLayers. Then I switched to a "vanilla" OpenLayers (without Vue) to test COG. The PR #261 was just the outcome of my tries. Maybe meanwhile, with a newer OpenLayers version, the upgrade problems can be solved easier.

pvgenuchten:
did a bit of research, the issue seems the use of older versions of webpack/babel in this project, because geotiff.js applies some recent js conventions. Updating webpack/babel seems not very obvious...

Issue #198, opened 2021-03-29
Point (De-)Cluster Renderer


JakobMiksch:
It would be useful to display vector data using a clustering or de-clustering visualisation.

Ideally configurable in the `appContext`. 

There are already OpenLayers based implementations from `ol-ext`
- [demo](https://github.com/Viglino/ol-ext/blob/master/examples/map/map.clustering.html)
- [cluster interaction](https://github.com/Viglino/ol-ext/blob/master/src/interaction/SelectCluster.js)

![image](https://user-images.githubusercontent.com/15704517/112858554-44560a80-90b2-11eb-9025-e91c5c124466.png)
![image](https://user-images.githubusercontent.com/15704517/112859123-d4944f80-90b2-11eb-95ee-772a50c0a849.png)

spwoodcock:
Would you be interested in a contribution for this?
Or would it be more valuable to work on #261 first?

fschmenger:
Hi @spwoodcock,
both of those would be valuable contributions, so it's definitely appreciated if you pick one and work on it.

Regarding #261 I'm not familiar with the latest state. Since we're in the process of moving towards Wegue 2 a migration to OL 7 would become feasible. But it's probably a good idea to move in small steps and support the latest OL 6 version first. 

Cheers Felix

chrismayer:
Hi @spwoodcock , thanks for your will to contribute. Very much appreciated! Recently we had a project where I did something in the direction of clustering / declustering. Unfortunately the clustering algorithm is not optimal in the current Wegue OL Version `6.4.3`. But it got optimized in later OL 6 versions. Therefore I'd suggest to go to latest OL v6 before working on a clustering solution.

spwoodcock:
Hi @chrismayer, I also have worked on a project with clustering on ol6.4+ and it worked nicely.

Let's wait for #320 then before starting this.

PR #334, opened 2023-04-27
Upgrade to OpenLayers v7

### LINKED ISSUE #329: Upgrade to latest OpenLayers
Wegue should support the latest (major) version ob OpenLayers, currently v7.x.

spwoodcock:
Fixes #329. I upgraded from ol v6.15.7 --> 7.2.0.

There were some minor deprecations to replace:
- Overlay no longer accepts `autoPanAnimation` as an option. Now the animation delay is passed as an object directly to the `autoPan` option.
- `map.forEachLayerAtPixel()` was removed. I replaced it with a `map.getLayers().forEach()`. This shouldn't be much of a performance issue as `getLayers` was called internally by `forEachLayerAtPixel` anyway. Perhaps there is a way to optimise, but I couldn't find it.
- I also updated a Vuetify deprecation that was giving a warning: `v-content` was simply renamed `v-main`.

Additional considerations:

- Internet explorer is now no longer supported by ol. I hope that's not a problem.
- With ol v7, now we can work on #198 and #290.


spwoodcock:
Merged `master` to only show specific commits - ready to test and merge.

spwoodcock:
Closed in favour of #335

PR #339, opened 2023-07-18
Extend MeasureTool by angle tool

chrismayer:
This extends the `MeasureTool` by a new functionality to measure/calculate an azimuth by drawing a 2-point line:

image

The function is hidden by default, so the `MeasureTool` looks as before in case anything is changed. To show / activate the functionality a new config property `showAngleTool` is introduced for the  `MeasureTool` module. So by setting this property to `true` in the `app-conf.json` you'll see the new angle tool.

```json
"wgu-measuretool": {
   "target": "toolbar",
   "win": "floating",
   "icon": "photo_size_select_small",
   "draggable": false,
   "strokeColor": "#c62828",
   "fillColor": "rgba(198,40,40,0.2)",
   "sketchStrokeColor": "rgba(198,40,40,0.8)",
   "sketchFillColor": "rgba(198,40,40,0.1)",
   "sketchVertexStrokeColor": "#c62828",
   "sketchVertexFillColor": "rgba(198,40,40,0.2)",
   "showAngleTool": true  <==
},
```


src/util/Angle.js
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Determine the angle between two coordinates. The angle will be between
+   * -180° and 180°, with 0° being in the east. The angle will increase
+   * counter-clockwise.
+   *
+   * Inspired by http://stackoverflow.com/a/31136507
+   *
+   * @param {Array<Number>} start The start coordinates of the line with the
+   *   x-coordinate being at index `0` and y-coordinate being at index `1`.
+   * @param {Array<Number>} end The end coordinates of the line with the
```
JakobMiksch, line 14: maybe here as type an OpenLayers coordinate can be used: https://openlayers.org/en/latest/apidoc/module-ol_coordinate.html#~Coordinate

src/util/Angle.js
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Determine the angle between two coordinates. The angle will be between
+   * 0° and 360°, with 0° being in the east. The angle will increase
+   * counter-clockwise.
+   *
+   * Inspired by http://stackoverflow.com/a/31136507
+   *
+   * @param {Array<Number>} start The start coordinates of the line with the
+   *   x-coordinate being at index `0` and y-coordinate being at index `1`.
+   * @param {Array<Number>} end The end coordinates of the line with the
```
JakobMiksch, line 37: maybe here as type an OpenLayers coordinate can be used: https://openlayers.org/en/latest/apidoc/module-ol_coordinate.html#~Coordinate

src/components/measuretool/MeasureResult.vue
```diff
@@ -1,51 +1,59 @@
...
   props: {
-    measureGeom: { type: Object }
+    measureGeom: { type: Object },
+    measureType: { type: String }
   },
   data () {
     return {
       area: ' -- ',
-      distance: ' -- '
+      distance: ' -- ',
```
JakobMiksch, line 31: the string ` -- ` appears quite often and could be extracted to an constant. But that's more a matter of taste ...

JakobMiksch (commented):
Code looks fine. Only minor remarks/ideas inline.

JakobMiksch (approved):
I did not run it. But the code looks fine. Can be merged from my side.

Issue #342, opened 2023-08-08
package-lock.json messed up since merge of #335


sronveaux:
Hi team,

While making some tests with latest commits of `Wegue` and implementing new things linked to #322 I stumbled across something messed up since merge of #335 and the commit 1cf2cfadbb2856b9d406ff8752f2be1d862831a1.

It seems like `package-lock.json` has been truncated and the last 11.000 lines are missing...

Due to the fact that it was regenerated with an older version of `NPM` (as stated in the docs) and that removing it completely and regenerating it from scratch adds some problems when running unit tests, I don't know which strategy would be the best to repair this or which one you'd prefer to follow...

Cheers,
Sébastien

chrismayer:
Closed by #344

PR #343, opened 2023-08-08
Corrected a Nominatim bug shown by unit tests

sronveaux:
Hi team,

Since last week, three unit tests have started to fail due to some change in `Nominatim API URLs`.

before

If you visit the `URL` written inside of the `OSM Geocoder provider`, some hints are given:

> Using the URL /search/ and /reverse/ (with slashes) is no longer supported. Please use URLs as given in the documentation.
> ...
> See [github issue #3134](https://github.com/osm-search/Nominatim/issues/3134) for more details.

This small PR corrects this small issue...

Cheers
Sébastien

Edit: As stated in #342, when package-lock.json is regenerated, some of the updated dependencies break a few more unit tests and it seems that `ci-tests` suffer the same problem as this PR works flawlessly at home but has problems when run on `Github` server.

after

This could be mitigated by configuring the build step to use `npm ci` instead of `npm i` as summarised, for example, in [this article](https://levelup.gitconnected.com/npm-i-vs-npm-ci-install-node-modules-in-your-app-faster-and-wisely-e5b1bef0f93d) :

> npm ci just installs existing dependencies, in contrast to npm install, which attempts to update current dependencies if possible. This ensures that the builds in continuous integration are reliable. It’s better to use npm i in development and npm ci for production.

### DIFF
--- a/src/components/geocoder/providers/osm.js
+++ b/src/components/geocoder/providers/osm.js
@@ -32,7 +32,7 @@ export class OpenStreetMap {
    */
   constructor () {
     this.settings = {
-      url: 'https://nominatim.openstreetmap.org/search/',
+      url: 'https://nominatim.openstreetmap.org/search',
       params: {
         q: '',
         format: 'json',


chrismayer:
The `package-lock.json` topic can be discussed in #342.

PR #345, opened 2023-08-18
Geocoding error handling

sronveaux:
Here's some kind of follow-up to #343 where a flaky behaviour was observed during Geocoder unit tests.  
Sometimes a strange error was thrown : `TypeError: this.results.forEach is not a function`

This happened from the `query` function inside the `GeocoderController` called from the `Geocoder.vue` component where the actual request was made like this:

```JavaScript
try {
  const response = await json(ajax);
  return this.provider.handleResponse(response)
} catch (err) {
  return err
}
```

But with this code, if the underlying `fetch` was in error, the exception was caught then **returned** which made the `Promise` from the `Geocoder component` **resolve** instead of **reject**. So the error was considered as a result but was an `Error` instance instead of the expected `Array` of results...

This was simply corrected by removing the `try - catch`. If an error is thrown, it is practically caught inside the `querySelections` method of the `Geocoder component` as planned first.

The other unit tests for the `geocoder` have been slightly refactored to use a fake request instead of doing multiple API calls to `Nominatim`.  
Some could argue that there is still a test making a real request and that this shouldn't be done inside unit tests but as this is the one which unveiled the bug corrected in #343 I thought making a real request was valuable in this case.

Please note that this final implementation is perhaps not the best one (`onQueryError()` does nothing except making a call to `trace()`) but the bug is corrected and the component should behave as it was designed at first...

### DIFF
--- a/src/components/geocoder/GeocoderController.js
+++ b/src/components/geocoder/GeocoderController.js
@@ -76,12 +76,8 @@ export class GeocoderController {
       ajax.callbackName = parameters.callbackName;
     }
 
-    try {
-      const response = await json(ajax);
-      return this.provider.handleResponse(response)
-    } catch (err) {
-      return err
-    }
+    const response = await json(ajax);
+    return this.provider.handleResponse(response)
 
     // // Do the query via Ajax XHR, returning JSON. Async callback via Promise.
     // json(ajax)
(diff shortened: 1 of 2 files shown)


chrismayer (approved):
Thanks for this @sronveaux. I just tested your change and this works quite well.

PR #346, opened 2023-08-21
Upgrade OpenLayers to v7.5.1

chrismayer:
This upgrades the OpenLayers dependency to version 7.5.1 (current latest).

Since OpenLayers follows semantic versioning no problems or conflict should appear when upgrading, which can be verified with the official change logs:

 - Change log [7.4.0](https://github.com/openlayers/openlayers/releases/tag/v7.4.0)
 - Change log [7.5.0](https://github.com/openlayers/openlayers/releases/tag/v7.5.0)
 - Change log [7.5.1](https://github.com/openlayers/openlayers/releases/tag/v7.5.1)


sronveaux (approved):
Just made some tests with my personal app and I can confirm everything works flawlessly.

It will be great to have support of WebGL Vector Layers (for example) available inside Wegue. This opens once again a lot of new possibilities !

Just a question though, would it be useful to add a link to [the full OpenLayers upgrade notes](https://github.com/openlayers/openlayers/blob/main/changelog/upgrade-notes.md) inside Wegue  upgrade notes so that users would have the opportunity to trace all backwards incompatible changes from `OL 6.4.3` to `OL 7.5.1` and perhaps later ?  
I know there is already a link to the release notes of V7 and perhaps this is enough...

This PR is indeed ready to merge from my point of view !

chrismayer:
Thanks for your review and feedback @sronveaux. I just added another commit, which adds a link to the  [the full OpenLayers upgrade notes](https://github.com/openlayers/openlayers/blob/main/changelog/upgrade-notes.md) as suggested by you. I totally agree that this is a very useful resource in an upgrade process.

I also gave this a shot in a bigger real-life application and still detected no problems...so I am going to merge now.

Thanks again!

PR #335, opened 2023-04-27
Update dependencies prior to Wegue v2

spwoodcock:
Updated any remaining non-dev dependencies, on top of all previous PRs in the V2 series.
Replaces #295.

I can't run the Karma tests easily on my system & don't have time for a workaround right now.
It would be worth running `npm run test` after all the updates.

Issue #322, opened 2023-03-13
Enforce Vuetify ESLint rules.


fschmenger:
As proposed within #312, we should consider installing the plugin [eslint-plugin-vuetify](https://www.npmjs.com/package/eslint-plugin-vuetify) to have access to the Vuetify ESLint rules. This should help us in maintaining vuetify code and prevent us from using outdated components, properties etc.

sronveaux:
Hi @fschmenger,

Indeed I was planning to work on this but waited for the `Wegue V2` branch to become a reality as all the PRs I've written until now are unfortunately not independent but cumulative. `V2` was branched from `master`, then `V3` was branched from `V2`, `V4` from `V3` and eventually `V5` from `V4`.

The `eslint-plugin-vuetify` require `ESLint7` or `ESLint8` and support of `ESLint7` was only added in `Vue-CLI V5`. It means working on it now would add a new level of dependency. So I'd like to tackle this once we have a branch including `Vue-CLI V5`.

I'd propose to do two or three linked things here:

1. Include the official [Vuetify plugin for Vue-CLI](https://www.npmjs.com/package/vue-cli-plugin-vuetify)
2. Include and enforce the `Vuetify ESLint rules`
3. Update `ESLint` from `V7` to `V8`

The last point can be done separately but should be useful to get up-to-date on this subject.  
`ESLint8` was not compatible with `Vue-CLI V4` but works well with `Vue-CLI V5` so I definitely have to give it a try!

fschmenger:
OK, thank you for pointing out the implications! I fully agree that we should put it on hold until #314 is finalized, as this will possibly result in a bunch of code formatting changes again. Again this isn`t urgent, so take your time!

sronveaux:
Hi @fschmenger,

This was addressed by #348 and #355.
Except if you think of something else, this one can be closed...

fschmenger:
Hi Sebastien, thanks a lot for implementing this and thanks to Christian as well for testing.
Sorry for being so quiet lately as I`m still very limited on time. Anyway I`m closing this issue for now.
Cheers Felix

PR #347, opened 2023-08-22
V2 upgrade docs and unit test specs files fix

sronveaux:
This is a very short upgrade towards `Wegue V2` which contains two `fixes`:

1. Some unit test files in the `test` folder were re-added when merging #335 but should be inside the `tests` directory. After checking, those files were simply doubles and can be safely removed.
2. Some upgrade notes concerning `Vue-CLI` and `ESLint` are added. The `ESLint` part should be amended once #322 will be implemented.


chrismayer (approved):
Thanks for sharing this very useful information, @sronveaux. Good to go. Please merge at will.

PR #352, opened 2023-10-18
Upgrade Vuetify to version 2.7.1

chrismayer:
Upgrades Vuetify to latests stable LTS version 2.7.1. I tested the dev-setup and the production build. No problems so far, which could expected when looking at the Vuetify changelog https://github.com/vuetifyjs/vuetify/releases/tag/v2.7.0 and https://github.com/vuetifyjs/vuetify/releases/tag/v2.7.1.

Part of #330

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "proj4": "2.9.0",
         "vue": "2.7.14",
         "vue-i18n": "^8.28.2",
-        "vuetify": "2.6.15"
+        "vuetify": "2.7.1"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


sronveaux (approved):
Thanks @chrismayer, it is very important, I think, to be on a LTS version of `Vuetify` on the upcoming next version of `Wegue`, so this is really appreciated!

Tested it in a bigger application and everything seems to work as expected. It looks perfectly fine for me.

chrismayer:
Thanks for your review @sronveaux,  going to merge now...

PR #354, opened 2023-10-18
Allow custom 'publicPath' config

chrismayer:
This introduces an ENV VAR `WGU_PUBLIC_PATH` in order to change the `publicPath` Vue CLI configuration, which is used to create the production build, without modifying the Wegue code base.

Also changes the default for `publicPath` from `'/'` to `'./'`, since the relative notation `'./'` works for most cases, e.g. if the app is deployed at the root of a domain and for a deployment under a certain path. For any other special case use the  `WGU_PUBLIC_PATH` as stated above.

See https://cli.vuejs.org/config/#publicpath

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,12 @@ npm run lint:fix
 
 See [Configuration Reference](https://cli.vuejs.org/config/).
 
+### ENV VARs
+
+Besides the environment variables supported by Vue CLI Wegue offers the following ENV VARs:
+
+- `WGU_PUBLIC_PATH` allows to modify the [publicPath](https://cli.vuejs.org/config/#publicpath) Vue CLI configuration, which is used in the production build. Default of `publicPath` is `'./'`.
+
 ## Run with Docker
 
 Versioned Docker images are available on [DockerHub](https://hub.docker.com/r/meggsimum/wegue/tags).
--- a/docs/workshop.md
+++ b/docs/workshop.md
@@ -1004,3 +1004,16 @@ npm run build
 ```
 
 Now, you can move the contents of the `dist` directory to a web space and it can be seen online.
+
+### Modify deployment path (publicPath)
+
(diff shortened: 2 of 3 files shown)


JakobMiksch (commented):
I am not sure how to test that. 
I ran `WGU_PUBLIC_PATH=foo npm run build ` , but it is not clear to me on which URL this is supposed to run.  I ran a webserver both inside the build directory and one level higher. In both cases I could not open the webmap.

sronveaux:
Hi @JakobMiksch,

The idea to test this here is to define the ENV_VAR inside a file called `.env.production` as explained in the [Vue-CLI official documentation](https://cli.vuejs.org/guide/mode-and-env.html#environment-variables).

By defining it in this file, you should be able to test this PR.

Regards,
Sébastien

sronveaux (approved):
Hi @chrismayer,

I made some tests here, defining the new environment variable and NOT defining it and managed to run a WebServer to make a complete app to run thoroughly in all cases. So yes, I think this works as expected.

However, for this new environment variable to be fully understandable, shouldn't it be documented somewhere ?  This is not easy as there are no documentation section concerning potential environment variables and all what is related to configuration is linked to the `app-conf.json` files, so I don't really see where it could fit nicely for now... perhaps, like what is done for application contexts, a little note inside the workshop in the [build for production part](https://wegue-oss.github.io/wegue/#/workshop?id=build-for-production) could do the trick ?

chrismayer:
Thanks for your feedback and testings @sronveaux and @JakobMiksch. I totally agree that a little documentation of the ENV VAR would be really helpful. So I added it to the README in the section [Production Build](https://github.com/wegue-oss/wegue#production-build) and to the workshop in the section [Build for production](https://wegue-oss.github.io/wegue/#/workshop?id=build-for-production) as suggested by @sronveaux. If necessary we could extend and improve those documentations.

Thnaks again. I am going to merge now...

PR #356, opened 2023-11-09
Introduce global app language lookup

chrismayer:
This PR introduces a global application language lookup `Vue.prototype.appLanguage`, which allows to access the current language in a static manner from normal JS-/ES- resources, without being in a Vue component.

### DIFF
--- a/app-starter/WguAppTemplate.vue
+++ b/app-starter/WguAppTemplate.vue
@@ -120,6 +120,7 @@ export default {
     }
   },
   created () {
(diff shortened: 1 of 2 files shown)


chrismayer:
@sronveaux You're totally right, I changed `appLanguage` to `$appLanguage` to follow Vue best practice.

Documentation of this global prop I would tackle in a separate step by generally documenting the global properties. With also taking some thoughts on which one should be usable by devs and which not (as already stated by you). I created an issue for that => #376

PR #357, opened 2023-11-09
Fix toQueryString function in UrlUtil for null values

chrismayer:
This PR fixes the `toQueryString` function in the `UrlUtil` in case a `null` value is passed in for a parameter. Main reason is that `typeof null` resolves to `'object'` (thanks JavaScript :roll_eyes: ), which leads to a recursion with null instead of an object.

### DIFF
--- a/src/util/Url.js
+++ b/src/util/Url.js
@@ -100,11 +100,13 @@ const UrlUtil = {
   toQueryString (obj) {
     return Object.keys(obj)
       .reduce((a, k) => {
-        a.push(
-          typeof obj[k] === 'object'
-            ? this.toQueryString(obj[k])
-            : `${encodeURIComponent(k)}=${encodeURIComponent(obj[k])}`
-        );
+        if (obj[k] !== null) {
+          a.push(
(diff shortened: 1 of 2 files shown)


fschmenger:
Hi Christian, looks good to me. One question here: Is it required to encode params as null or should they just be removed from the query string?
If the answer is 'yes' then feel free to merge.

chrismayer:
Thanks for your review @fschmenger. I don't think it is required to encode URL params as null and maybe a bad idea if I think twice. There might be typing problems in some API implementations receiving `foo=null`, since `null` is then a string and might not match a defined data type.

So I decided to rework this a bit and remove null values from the URL query string. Would you please give this another quick check, @fschmenger ? If you are OK with the code please hit the `Approve` button, otherwise the PR is not mergable... Thanks in advance.

fschmenger:
Ok, looks good.

fschmenger (approved):

chrismayer:
Thank you @fschmenger :pray: Going to merge now...

Issue #351, opened 2023-10-18
Upgrade to Vuetify 3


chrismayer:
Due to EOL of Vue 2 an upgrade of Vue to version 3 will come up. This would force us to switch to Vuetify 3 even tough its LTS period is a bit longer (January 23rd, 2025) than for Vue 2.

Should be done after or in combination with #350.

sronveaux:
Hi team,

As explained in #350, I'd personally tackle this one after it as `Vuetify3` still lacks some of the functionalities that were present inside `Vuetify2`. It even seems that some people changed their mind and used, for example, `Quasar` instead of `Vuetify` because of that. (For example, see [this discussion on official Vuetify Github](https://github.com/vuetifyjs/vuetify/issues/13518)

JakobMiksch:
meanwhile Vuetify3 re-adds more and more components, that were present in Vuetify2 and were currently marked as experimental (see 3.4 release https://vuetifyjs.com/en/getting-started/release-notes/#user-content-release-notes)

Issue #350, opened 2023-10-18
Upgrade to Vue 3


chrismayer:
Since "Vue 2 will reach End of Life (EOL) on December 31st, 2023" (see https://v2.vuejs.org/lts/) we should tackle the upgrade of the code base to Vue 3. Even tough offering different new concepts in Vue 3 one could stay at the classic component approach of Vue 2. So the upgrade effort should be OK besides the normal upgrade madness, but no complete re-write.

sronveaux:
Hi team,

I have to tell I already looked at this last year as this was some sort of natural follow-up of migrating to the latest `Vue-CLI` but what stopped me was the lack of compatibility of `Vuetify` at that time.

From a completely personal point of view, I'm not looking at it for now neither because `Vuetify3` still lacks some components which were available in V2. This would force me to stop upgrading `Wegue` inside my main project which unfortunately makes use of some of those.

After some reading though, it seems there can be a possibility to upgrade `Vue` only using the `compat migration build` which could potentially still work with `Vuetify2`. The migration build could then be entirely replaced by `Vue3` once `Vuetify` will be complete enough to be migrated too...

For the rest, as @chrismayer said, we should be able to stay with the `Options API` and migrate without too much trouble, the main difficulty will be to replace the `WguEventBus` as this way of working is not possible as is anymore. There exists some possibilities to reach the same behaviour like the excellent [mitt](https://www.npmjs.com/package/mitt) for example.

JakobMiksch:
> the main difficulty will be to replace the `WguEventBus`

in Vue 3 it is possible to store state like a map in composables or with the "official" state management library "pinia"

JakobMiksch:
For sharing the map state across components in Vue 3, I find composables quite useful. 
Here a library with included demo project, that could be of interest of the next generation of Wegue: 
- Repo: https://github.com/JakobMiksch/vue-ol-comp
- Demo App: https://jakobmiksch.github.io/vue-ol-comp/
- npm: https://www.npmjs.com/package/vue-ol-comp

fschmenger:
@sronveaux Great work. I hopefully will get to test the Vue3 migration in a custom app soon. Should we close this issue and possibly #351 ?

sronveaux:
Hi @fschmenger, thanks for the comment! I hope this will suit your expectations and that the migration won't be too complicated. Don't hesitate to give feedback on the process afterwards!

[...]

PR #360, opened 2023-11-17
Changed deprecated slot syntax

sronveaux:
In the wake of #359 there is a small PR which gets rid of the `slot` and `slot-scope` attributes which are strongly discouraged since Vue 2.6 and deprecated in Vue 3.

This PR was made as a small step towards #350 and #351 and is in fact the preliminary step before migrating to Vue 3 as stated in the [official migration guide](https://v3-migration.vuejs.org/migration-build.html#preparations).

This should have no impact on any working Wegue application as deprecated syntax can still be used next to the new one but it should be tested on bigger and more complete apps just to be sure...

### DIFF
--- a/app-starter/WguAppTemplate.vue
+++ b/app-starter/WguAppTemplate.vue
@@ -5,11 +5,21 @@
 
     <wgu-app-header>
       <!-- forward the slots of AppHeader -->
-      <slot name="wgu-tb-start" slot="wgu-tb-start" />
-      <slot name="wgu-tb-after-title" slot="wgu-tb-after-title" />
-      <slot name="wgu-tb-before-auto-buttons" slot="wgu-tb-before-auto-buttons" />
-      <slot name="wgu-tb-after-auto-buttons" slot="wgu-tb-after-auto-buttons" />
-      <slot name="wgu-tb-end" slot="wgu-tb-end" />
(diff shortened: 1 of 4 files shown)


fschmenger:
Hi @sronveaux, thanks a lot for providing this! I`m unfortunately a bit short currently for a full review but will do either in the upcoming weeks or maybe @chrismayer can?!

Do the commits 607c6a3373d5fa62918914ad1ce25c549b815903 and 49ea5f30f7c666397e25ceeba83b18a45df42a9c relate to slot syntax changes?
Maybe you should either change the title of this PR to something more general like "maintenance" or rather split it up into 2 separate PRs (can be nested). 

Greets Felix

sronveaux:
Hi @fschmenger and thanks for the comment, no problem, there is no hurry, this will be useful later when updating to `Vue3` but it has no other purpose...

Concerning the two commits, it was kind of linked because when updating the workshop documentation with the new syntax, I came across the fact that the workshop code couldn't be run anymore because of linting errors. So it was no point to change just the slot syntax but have a code that can't be run on current version of Wegue...

No problem, this PR now just changes the slot syntax in code and documentation and I'm going to create another PR with the updates to the workshop...

Greets, Sébastien

PR #362, opened 2023-12-01
Unit test infrastructure upgrade

sronveaux:
This upgrades Chai to version [4.3.10](https://github.com/chaijs/chai/releases) and Sinon to version [17.0.1](https://github.com/sinonjs/sinon/blob/main/CHANGES.md#1701).

Karma plugins are also updated to latest available versions. This concerns `karma-sourcemap-loader` and `karma-spec-reporter`.

Unit tests are working as expected on Wegue codebase and on some other projects too.


sronveaux:
Hi @fschmenger,

Thanks for reviewing this one.

To answer your question, all were updated to the latest version published at the time this PR was written. You can see in the following snapshot of a run of `npm outdated` that sinon-chai or karma-xxx dependencies are not listed (except karma-webpack which has a new version published after this PR was released).

sinon-chai is requested with `^3.5.0` inside package.json but has `^3.7.0` installed in package-lock.json. This is the latest version pubished 3 years ago now.

image

I can amend this PR and set versions reported as installed inside package-lock.json to package.json though if you wish...

Two last things concerning this for the future:

- Chai has a new major version published since this PR was written. However, I'm 99% sure Karma would be incompatible with V5.x because of ESM-only support (where Karma adds its plugins through `require()` calls).
- karma-webpack has a new V5.0.1 which seems has introduced breaking changes like, for example, requiring Node >= V18.x where Wegue currently requires Node >= V16.19.0

With those things in mind, I'd personally stick with current PR for now and see afterwards whether we can update the unit test infrastructure once again... also keeping in mind that karma itself was flagged as deprecated in April 2023...

fschmenger:
Hi Sebastien,
good to hear from you and thanks for the explanation.
Regarding karma-webpack: It's always a pain when they introduce breaking changes within a minor version update and there`s no point going through all the headaches right now.
So I agree, we can look into this later on when we possibly have to move away from Karma anyway.

I will merge then.

PR #363, opened 2023-12-01
Changed deprecated syntax in .eslintrc test file

sronveaux:
This simple PR corrects the content of the `.eslintrc` file which is used while generating unit test builds.

The `globals` definition there was still using [an older deprecated syntax](https://eslint.org/docs/latest/use/configure/language-options#using-configuration-files-1).


fschmenger (approved):
Looks fine to me, I`m going to merge.

PR #364, opened 2023-12-01
Enable Vuetify treeshaking

sronveaux:
This PR is a proposal to enable [Vuetify treeshaking](https://v2.vuetifyjs.com/en/features/treeshaking/#treeshaking).

This simple tweak reduces the total size of the `dist` directory of `Wegue starter app` from 18.1MB to 14.5MB.

Some technical difficulties are encountered with unit tests though once the `Vuetify loader` is installed. Please note that this issue also happens on a fresh app scaffolded with `Vue-CLI`. The way they get around this when generating an app from scratch is to put `vuetify` in the `transpileDependencies` list, which is currently empty in `Wegue` by default.  
This transpile operation makes the bundle size to grow of 26.374 bytes with the starter app which is negligible, but also produces a warning: `[BABEL] Note: The code generator has deoptimised the styling of C:\Sources\Wegue\node_modules\vuetify\dist\vuetify.js as it exceeds the max of 500KB.`

The only other way around I've found is to disable treeshaking once `NODE_ENV === 'test'` inside `vue.config.js`. I haven't encountered problems in my personal applications with this yet but this should be tested on more apps I think...

So this PR implements the second solution (disabling treeshaking during unit tests) but in case you'd like to test the other option, all what should be done is commenting out line 65 in `vue.config.js` and replacing line 75 of the same file by

```js
  transpileDependencies: ['vuetify']
```

Just tell me whether treeshaking would be of an interest and what you think about this,

Cheers
Sébastien

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1,7 +1,7 @@
 // The Vue build version to load with the `import` command
 // (runtime-only or standalone) has been set in vue.config.js with runtimeCompiler.
 import Vue from 'vue';
-import Vuetify from 'vuetify';
+import Vuetify from 'vuetify/lib/framework';
 import PortalVue from 'portal-vue'
 import VueI18n from 'vue-i18n';
 import 'roboto-fontface/css/roboto/roboto-fontface.css'
@@ -13,7 +13,6 @@ import UrlUtil from './util/Url';
 import LocaleUtil from './util/Locale';
 import ObjectUtil from './util/Object';
 import ColorThemeUtil from './util/ColorTheme'
-import 'vuetify/dist/vuetify.min.css';
 import axios from 'axios';
 
 Vue.use(Vuetify);
--- a/vue.config.js
+++ b/vue.config.js
@@ -57,10 +57,12 @@ module.exports = defineConfig({
     // Tweak configuration options for Karma test runner to produce a bundle
     // which can run under Chrome headless. Avoid warnings due to custom entries
     // and customized filenames. Enable correct code coverage of .vue files.
+    // Disable Vuetify treeshaking.
     if (process.env.NODE_ENV === 'test') {
       config.devtool = 'eval'
       config.optimization.runtimeChunk = false
       config.optimization.splitChunks = false
+      config.plugins = config.plugins.filter(plugin => plugin.constructor.name !== 'VuetifyLoaderPlugin')
       delete config.target
       delete config.entry
       delete config.output.filename
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,7 +56,8 @@
         "sinon": "^9.2.4",
         "sinon-chai": "^3.5.0",
         "vue-cli-plugin-vuetify": "~2.5.8",
-        "vue-template-compiler": "2.7.14"
+        "vue-template-compiler": "2.7.14",
+        "vuetify-loader": "^1.9.2"
       },
       "engines": {
         "node": ">= 16.19.0",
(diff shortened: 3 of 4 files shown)


src/main.js
```diff
@@ -1,7 +1,7 @@
 // The Vue build version to load with the `import` command
 // (runtime-only or standalone) has been set in vue.config.js with runtimeCompiler.
 import Vue from 'vue';
-import Vuetify from 'vuetify';
+import Vuetify from 'vuetify/lib/framework';
```
fschmenger, line 4: In the documentation from https://v2.vuetifyjs.com/en/features/treeshaking/#vuetify-loader they import vuetify from 'vuetify/lib'. Is this something that you changed or is this the default provided by the template? I understand that this probably doesn`t make any difference to bundle sizes after treeshaking - see https://stackoverflow.com/questions/67615517/vuetify-difference-between-import-vuetify-lib-framework-vs-vuetify-lib. Is there any performance difference regarding build time?

vue.config.js
```diff
@@ -57,10 +57,12 @@ module.exports = defineConfig({
     // Tweak configuration options for Karma test runner to produce a bundle
     // which can run under Chrome headless. Avoid warnings due to custom entries
     // and customized filenames. Enable correct code coverage of .vue files.
+    // Disable Vuetify treeshaking.
     if (process.env.NODE_ENV === 'test') {
       config.devtool = 'eval'
       config.optimization.runtimeChunk = false
       config.optimization.splitChunks = false
+      config.plugins = config.plugins.filter(plugin => plugin.constructor.name !== 'VuetifyLoaderPlugin')
```
fschmenger, line 65: I think disabling treeshaking for unit tests is a perfectly viable solution. I guess that this will also be faster when building the test-bundle.

fschmenger (approved):
Hi @sronveaux,
I tested this in a larger application and it worked out fine.
I can also confirm your observations on Wegue default app bundle sizes.
For my custom app it`s roughly 18,6 MB with and 21,0 MB without treeshaking.

Nice work!
One question inline, otherwise approved and ready to merge.

src/main.js
sronveaux: Hi @fschmenger, concerning this one, the template generated when adding `Vuetify` to a plain `Vue` app makes use of the `vuetify/lib/framework` indeed which is fun as the template was written by `Vuetify` team... so it seems they did things differently in their docs and examples...

So that's the simple reason why I sticked to this one when I wrote this PR. Moreover, the contents of the file `vuetify/lib/index.js`, is the following:

```javascript
export * from './components';
export * from './directives';
export { default as colors } from './util/colors';
export { default } from './framework';
//# sourceMappingURL=index.js.map
```

which means that in our case, the result is the same as we import the default. This would be different if we wanted to import components separately as explained [here](https://v2.vuetifyjs.com/en/features/treeshaking/#manually-importing).

To further report on your comment, I just did some performance tests and I must admit I don't really understand all the results...

At first, I'm unable to reproduce the facts about bundle size that is explained in the very interesting link to stackoverflow you provided. Forcing bundle in development mode, here is the result while importing `framework` directly :

image

and here is the results of importing from `lib` :

image

As we are now using the V2.7.X version of Vuetify, I suppose they changed something compared to what was observed in the SO post which should have been done with V2.5.X or V2.6.X... 

Now the interesting part is that I ran the build step multiple times, in development and production modes, with the import from `lib` and from `lib/framework` and I clearly see that building while importing from `framework` can get faster in dev and in prod modes !  Where I get a mean of 42 seconds to build when importing from `lib`, it gets down to 36 seconds when importing from `framework` !  The funny point is that it goes 15% faster to give the exact same result !  I suppose the import from `lib` creates a dependency graph which is more complicated but resolves the same in the end...

fschmenger: Yeah, this makes sense. Probably the 3 extra imports from `lib` will just be redundant when using defaults and eventually the treeshaker has to eliminate them again?!
Anyway, thanks for looking into this. I think since this is the template default and also gives better build performance we stick with it and come back to it later, if there is any requirement to do so.
I will merge then.

PR #366, opened 2024-02-26
Global feature hover options

fschmenger:
This draft adds a top level option `mapHover` to app.conf, to provide application wide defaults for feature hover customization. The following properties are supported (quoted from the doc):

| Property           | Meaning   | Example |
|--------------------|:---------:|---------|
| delay              | Timespan in milliseconds, by which displaying the tooltip is deferred after the mouse pointer rests. Defaults to `150`. | `"delay": 150`
| hideOnMousemove    | Hide the tooltip when the mouse cursor is moved. Defaults to `false`. | `"hideOnMousemove": false`
| hoverOverlay       | ID of a custom map overlay to use as a default display when a feature of the layer is hovered. Declaration of the `hoverOverlay` property on a mapLayer level takes precedence. For more information on how to implement a map overlay see the [reusable components](reusable-components?id=map-overlay) section. Defaults to Wegue's default tooltip `wgu-hover-tooltip` |  `"hoverOverlay": "my-custom-overlay"`

Example:

```json
"mapHover":
    {
      "delay": 150,
      "hideOnMousemove": false,
      "hoverOverlay": "my-custom-overlay"
    }
```

Additionally, some problems related to dangling tooltips have been fixed when the mouse cursor leaves the map canvas (efacbde3d8a6a2c04c5ae466232f235c4ac37ef7).

### DIFF
--- a/docs/wegue-configuration.md
+++ b/docs/wegue-configuration.md
@@ -19,6 +19,7 @@ This describes the Wegue application configuration, which is modelled as JSON do
 | **mapCenter**      | Initial center of the map in map projection | `"mapCenter": [0, 0]` |
 | mapProjection      | Configuration object for CRS / projection used for the map | see [mapProjection](wegue-configuration?id=mapprojection) |
 | mapGeodataDragDop  | Configuration object for geodata file drag/drop functionality on the map. Only by setting the config this function will be enabled. | see [mapGeodataDragDop](wegue-configuration?id=mapGeodataDragDop) |
+| mapHover           | Configuration object to customize the bevahior of hover tooltips on the map. | see [mapHover](wegue-configuration?id=mapHover)
 | **modules**        | Array of module configuration objects | See [modules](module-configuration) |
 | **mapLayers**      | Array of map layer configuration objects | See [mapLayers](map-layer-configuration) |
 | overviewMap        | Configuration object for the overview map. | See [overviewMap](wegue-configuration?id=overviewMap)
@@ -225,6 +226,29 @@ Below is an example for such a configuration object:
       "displayInLayerList": true
     }
 ```
+
+### mapHover
+
+Optional configuration object to customize the behavior of hover tooltips on the map. 
+The following properties are supported:
+
+| Property           | Meaning   | Example |
+|--------------------|:---------:|---------|
+| delay              | Timespan in milliseconds, by which displaying the tooltip is deferred after the mouse pointer rests. Defaults to `150`. | `"delay": 150`
+| hideOnMousemove    | Hide the tooltip when the mouse cursor is moved. Defaults to `false`. | `"hideOnMousemove": false`
+| hoverOverlay       | ID of a custom map overlay to use as a default display when a feature of the layer is hovered. Declaration of the `hoverOverlay` property on a mapLayer level takes precedence. For more information on how to implement a map overlay see the [reusable components](reusable-components?id=map-overlay) section. Defaults to Wegue's default tooltip `wgu-hover-tooltip` |  `"hoverOverlay": "my-custom-overlay"`
+
+Example:
+
+```json
+"mapHover":
+    {
+      "delay": 150,
+      "hideOnMousemove": false,
+      "hoverOverlay": "my-custom-overlay"
+    }
+```
+
 
 ### lang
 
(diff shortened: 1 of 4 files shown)


fschmenger:
Hi guys. This is a first draft. Some topics to think about:

We now have 3 spots to configure feature hover tooltip behavior: 1. (Optional) App-wide defaults, 2. settings at mapLayer scope and 3. (optional) properties to configure the MapOverlay when doing a custom implementation. Is this OK or could it be generally confusing? To explain the 3 config levels and my motivation. 
* Apparently `delay` cannot be configured on a layer level, since the pointer rest timer must be known in advanced before issuing any WMS GetFeatureInfo requests.
* I then added the `hoverOverlay` property for convenience, this can be declared app-wide or on a layer level.
* In our testing, the hover overlays turned out to be pretty disturbing when hovering larger areas of connected features, e.g. a complete tile layer. Therefore I added `hideOnMousemove` which e.g. can be used with a somewhat larger delay to improve the situation. I only added this to the app-wide setting, but it could be introduced on a layer level as well.

Most of the new property names have been ad-hoc choices. Any suggestions for renamings are welcome. E.g. should it be something like `mouseRestDuration `instead of `delay`, or something like `featureHover` or just `hover` instead of `mapHover`?

Let me know what you think.

docs/wegue-configuration.md
```diff
@@ -19,6 +19,7 @@ This describes the Wegue application configuration, which is modelled as JSON do
 | **mapCenter**      | Initial center of the map in map projection | `"mapCenter": [0, 0]` |
 | mapProjection      | Configuration object for CRS / projection used for the map | see [mapProjection](wegue-configuration?id=mapprojection) |
 | mapGeodataDragDop  | Configuration object for geodata file drag/drop functionality on the map. Only by setting the config this function will be enabled. | see [mapGeodataDragDop](wegue-configuration?id=mapGeodataDragDop) |
+| mapHover           | Configuration object to customize the bevahior of hover tooltips on the map. | see [mapHover](wegue-configuration?id=mapHover)
```
sronveaux, line 22: We can say this is simple detail level, but to keep the description in the same style as the other properties of the general section, wouldn't it be better to write it as `Configuration object for hover tooltips behavior on the map` or, like the `legend` one, `Configuration object containing application wide parameters for hover tooltips` ?

docs/wegue-configuration.md
```diff
@@ -226,6 +227,29 @@ Below is an example for such a configuration object:
...
+### mapHover
+
+Optional configuration object to customize the behavior of hover tooltips on the map. 
+The following properties are supported:
+
+| Property           | Meaning   | Example |
+|--------------------|:---------:|---------|
+| delay              | Timespan in milliseconds, by which displaying the tooltip is deferred after the mouse pointer rests. Defaults to `150`. | `"delay": 150`
+| hideOnMousemove    | Hide the tooltip when the mouse cursor is moved. Defaults to `false`. | `"hideOnMousemove": false`
+| hoverOverlay       | ID of a custom map overlay to use as a default display when a feature of the layer is hovered. Declaration of the `hoverOverlay` property on a mapLayer level takes precedence. For more information on how to implement a map overlay see the [reusable components](reusable-components?id=map-overlay) section. Defaults to Wegue's default tooltip `wgu-hover-tooltip` |  `"hoverOverlay": "my-custom-overlay"`
```
sronveaux, line 239: Again a detail here, I personally think "when a feature of **the** layer is hovered" is a bit inaccurate.

It would be more correct with something like "...when a feature of **a hoverable layer** is hovered" which is not really great as we have two times the **hover** word very close together.  
I first thought about "...when a feature of **a hoverable layer** is under the mouse pointer" which once again becomes vague and perhaps too verbose... if you have a better idea...

fschmenger:
Hi @sronveaux,
thanks for the review and the valuable feedback. I will go through the points you listed above.

Regarding `hideOnMousemove`: This could be a property on the layer in addition to the global setting. So the global setting would be just a default as well. However this probably would require a different name on the layer level.

Anyway, since this isn`t urgent, we can wait until @chrismayer is back, so he can have a word about it too.

docs/wegue-configuration.md
fschmenger: This also had a bad typo. I chose your 2nd formulation :)

sronveaux: Thanks for thinking about it !

docs/wegue-configuration.md
sronveaux: Good job, this will be clearer for end users as you wrote it !

sronveaux (approved):
Hi @fschmenger,

Sorry for my late reply, all is much clearer now with documentation amended like this.

I approve for merge, except if @chrismayer has something else to add ?

chrismayer (approved):
This looks really good and I am totally fine with this. Thanks a lot for providing this enhancement @fschmenger and for your feedback @sronveaux :+1: 

@fschmenger Please feel free to merge this PR.

Issue #365, opened 2024-02-22
OL layer properties are no longer reactive


fschmenger:
It seems that lately vue-bindings to `layer.getProperties()` or via the `layer.get('someProperty')` will not get updated. Particularly puzzling is, that it seems to work for all layers added in the 1st generation, i.e. those configured statically in app.conf but not for those added dynamically in a later cycle.

To see an example: Create a drag drop layer (e.g. download GeoJSON from https://openlayers.org/en/latest/examples/drag-and-drop.html) and open the layer list module. Then switch the active language. The drag drop layer is localized and supposed to switch its language, however only the "static" wegue layers will receive the update.

The related code in `LayerListItem` looks like
```
<v-list-item-title>
        {{ layer.get('name') }}
</v-list-item-title>

```
Neither does it work via `getProperties()` if you alter the code above to
```
<v-list-item-title>
        {{ layer.getProperties().name }}
 </v-list-item-title>
```
or by using explicit computed properties, etc.

Note that some other components, i.e. the AttributeTable will still update the layer name. This is just a happy circumstance because they use other localized attributes in the same block, causing the whole thing to re-render.

Trying to track it down I rolled Wegue all the way back to v1.2.1 to finally see it working. However, if I delete the package-lock.json and do a full reinstall of dependencies, then the functionality is broken even back then.

A workaround could be to register for layer `propertychange` events and then manage our own binding targets, e.g.
```
layer.on('propertychange', (evt) => {
   if (evt.key === 'name') {
       this.layerName = layer.get('name')
  }
});
```
We used to have something like this in a very old version of Wegue but it would be cumbersome to reintroduce it and hunt down all code blocks where a binding to layer properties occurs.

Any ideas on the underlying cause and what to do about it?

sronveaux:
Hi @fschmenger,

I tried to have a look at this one and arrive at the same results as you for now...

To give some more details of what I saw, if you install dependencies on Wegue v1.2.1 then remove package-lock.json and try to recreate it, the problem doesn't happen neither. You have to recreate the package-lock.json with an empty node_modules directory for the problem to occur. In this situation, the number of changed dependencies is so big that tracking what could induce this could potentially take ages...

Another funny thing is that when you reproduce the bug with the drag and drop as you explained, the name is correctly regenerated once you hide or show the layer by clicking on the chekbox next to the name... then if you switch languages again, you once again have to hide or show it for the name to be regenerated...

[...]

PR #368, opened 2024-03-18
Fix layer property reactivity

fschmenger:
This is my take on how we could possibly fix the layer-property-reactivity issue mentioned in #365.
To keep changes minimal I decided to wrap up OL layers and OL layer collections by proxy objects. The proxy objects behave mostly transparent and forward invocations to the underlying layer / collection objects, so they can be basically used as a drop-in.  Apparently, use them only for Vue components when reactivity is required.

Using LayerProxy around an OL layer:
* Construct a proxy object: `layerProxy = new LayerProxy(myLayer)`
* Proxy objects must be destroyed before they go out of scope: `layerProxy.destroy()`
* Use the underlying layer to pass it on to child components and APis: `layerProxy.toRaw()`
Internally, the layer proxy traps the `getProperties()` and `get()` methods to return a self managed object of key value pairs, which is synced via OL observables.

Similarly, using LayerProxyCollection around an OL Collection:
* Construct a proxy object: `collectionProxy = new LayerCollectionProxy(myLayerCollection);`
* Proxy objects must be destroyed before they go out of scope: `collectionProxy.destroy()`
* Use the underlying collection to pass it on to child components and APis: `collectionProxy.toRaw()`
Internally, the collection proxy traps the `forEach()`, `item()` and `getArray()` methods to return LayerProxy objects.

Currently the application uses only a single instance of `LayerProxyCollection`, which is managed inside `composables\Map`. The `useMap` function which formerly returned a list of layers has been altered to return the `LayerProxyCollection`. As a result instances of `LayerProxy` objects are also shared between components to make this as resource friendly as possible.

However constructing additional `LayerProxyCollection ` and  `LayerProxy` objects is feasible if it should be required by some components (although I dont see yet why this should be necessary).


fschmenger:
Hi guys, here is where to take a closer look:
* Are we happy with the overall solution? I couldn`t come up with something simpler but maybe there is...
* LayerProxy currently only manages properties, which are passed into the constructor. Other properties are not accessible. Is this a good design or should we try and manage/sync all available properties on the layer? (However those would have reactivity restrictions as described in #365)
* Unfortunately proxy objects must be destroyed to free event listeners and there is no JS destructor syntax either. Currently I did not implement reference counting. So the assumption is that you only destroy Proxy objects that you have created. E.g. don`t destroy LayerProxy objects which are maintained by the LayerProxyCollection.
* This is why LayerCollectionProxy currently does not trap and handle the pop(), push(), remove(), removeAt() and setAt() methods and just accepts/ returns raw OL layers there- see [OL docs](https://openlayers.org/en/latest/apidoc/module-ol_Collection-Collection.html). I didn`t wanna reason about ownership patterns for now.


Unfortunately I will have to focus on other projects again in the upcoming months, so take your time for review.
If we're happy with the overall solution, feel free to work on it. It should`t be too much from here on.

Cheers Felix

fschmenger:
Hi everyone, I’ve just updated this PR.

With the changes in Vue 3, the reactivity system has improved significantly, particularly with the introduction of proxies. However, the core issue remains: Vue’s reactivity doesn't track changes made to objects unless those changes go through its reactive proxy.

In our case, OpenLayers layers are managed internally by the OpenLayers API, which means we can't guarantee that all changes to those layer objects will be made via Vue’s reactive system. To work around this limitation, I’ve introduced proxy objects that wrap the layer instances. This allows us to better integrate with Vue's reactivity while preserving compatibility with OpenLayers.

The implementation should be feature complete:
* Proxy objects have been fully migrated for Vue 3 compatibility.
* It's no longer necessary to explicitly list properties to be tracked — the layer proxy now observes all property changes and dynamically tracks additions and removals.
* The Map composable has been updated to use a fully reactive `LayerProxyCollection`, so the required changes to consuming components are minimal.

Some of the components still use the raw OL layers, e.g. Geolocator, as they don`t require any reactivity and are not obtaining their layers via Map composable. Do we see any need to change those?

For testing: To see an example, where reactivity failed previously: Open the layer list module and switch the language.

Cheers Felix

fschmenger:
I decided to go with approach 2 and map all getter functions declared in `ol/layer/Base` to the internal reactive properties object. On a 2nd thought doing it in various controls like `LayerList` etc. does not improve the situation and only has the risk of having to fix code scattered across the whole project. I hope you agree :)

One last question: Is `ReactiveLayer` a better name than `LayerProxy`?

fschmenger:
Since this PR has been around for quite some time, a friendly reminder whether you could possibly do a review on it? :)
I think it is a bug with higher severity as several layer based controls (especially the layer list) will not update properly.

Note: I have some test code to fiddle around with with layer properties interactively. The layers can be accessed as OL raw layers and also as proxied layers. So you can see the differences in reactivity. Let me know if you need it.

sronveaux:
Hi @fschmenger,

Just a message to let you know that I do keep it in mind and followed thouroughly what you did on this one. Reviewing it is definitely on my Wegue project to-do list but as what is embedded has such a big impact (with custom proxies being created, potentially interacting with Vue reactivity internals), I really need a big time slot where I can fully immerge in it, think about it and fully test it.  
I'm already sure what you did is a solution, I just couldn't find a long enough time slot following migration to Vue 3 (which could also have implications on this one) due to the other project I have to work on here.

Once again, be sure the idea is not to pass it up or miss out...

PR #370, opened 2024-03-21
wfs endpoint with param fix

enricofer:
Hi, this PR fixes an issue in WFS URL formatting that prevent connection to wfs endpoints urls that already contain parameters, like qgis server connections: [http://qgisserver?MAP=/path/to/project&](http://qgisserver/?MAP=/path/to/project&)


chrismayer (approved):
LGTM, thanks @enricofer.

PR #375, opened 2024-04-12
Explicitly call 'npm run lint' in CI test pipeline

chrismayer:
This starts the linting process (`npm run lint`) explicitly in the CI test pipeline because the implicit lint call in the `npm run test` command does not break the pipeline, even if there are linting errors. So possible linting errors land in the Wegue code base if not discovered by human reviewers, which leads to a broken build.


JakobMiksch (approved):
LGTM

chrismayer:
Thanks for the review @JakobMiksch.

PR #377, opened 2024-04-18
Update ESLint to version 8.57.0

sronveaux:
This PR updates ESLint 8.57.0.

All plugins are also updated and some which were referenced inside `package.json` but are now included as dependencies of other packages were removed to keep direct dependencies list as clean as possible.

This PR is also the opportunity to write some comments about ESLint future updates so everyone is aware of it :

1. Since V8.53, ESLint team decided to deprecate all formatting rules. More info can be found on [this official announcement](https://eslint.org/blog/2023/10/deprecating-formatting-rules/).  
To keep it simple, those rules will stay at least in version 9 of ESLint but will not receive any updates anymore. Those rules should be abandoned in favor of a code formatter OR replaced by the rules defined in `[@stylistic/eslint-plugin-js](https://eslint.style/packages/js)`. This package contains all the deprecated rules and will be maintained. However, as our main dependencies (`eslint-plugin-vue` and `eslint-config-standard`) don't make use of those rules yet, it is better to wait before making the transition on our side...

2. A new format for ESLint configuration files is supported since 2022 and will be the default one in version 9. Transitioning to this new configuration file format should be done but once again when our main dependencies will export those kind of files. This is already the case for `eslint-plugin-vue` but not for `eslint-config-standard` (well it is already if you checkout the master branch on GitHub but not if you install the official NPM package).  
Making the transition will force us to get rid of `@vue/eslint-config-standard` which is used automatically by `vue-cli` and first versions of `create-vue`. What it basically does is configure ESLint to use `vue` rules and `standardJS` ones.  
Newer versions of `create-vue` don't use it anymore but don't include `standardJS` anymore neither so I think we could easily depart from `@vue/eslint-config-standard` but what about `standardJS`?  Should we keep it? I personally think we should as I like this style guide but please give me your opinion on this one!

Last point, it is still unsure whether ESLint V9 is compatible with `Vue-CLI` so perhaps all those thoughts will only be of interest when `Wegue` will be upgraded to `Vue3`...


chrismayer:
Thanks for this update @sronveaux and the detailed explanations. I like using up-to-date dependencies and the fact that you reduce explicit sub-dependencies.

I created separate issues for your points 1. #380  and 2. #379. The decisions can or have to be easily postponed and the corresponding discussions can be done in the respective issue. So this PR is not blocked from merging and things keep a bit more cleaner. 

So I am +1 to merge this.

chrismayer (approved):

sronveaux:
Thanks @chrismayer,

I agree 100% with you on this one. We have time to do those changes as ESLint has announced that those deprecations will not be removed for a long time...

I'll keep an eye on the "blocking" dependencies and are open to all what will be proposed in the two issues you opened!

PR #381, opened 2024-04-21
Print module

enricofer:
Hi, I'm submitting a new PR regarding a new print feature that allows to export the current map to pdf.
The implementation is adapted from the following Openlayers example: https://openlayers.org/en/latest/examples/print-to-scale.html
I hope you find of some interest for the project. In this case please point me out how to develop and integrate continuous testing for this module.

All the best.


chrismayer:
Thanks for your PR @enricofer! 

Your branch has conflicts and it seems there are changes of your previous PRs in the change sets, which makes it hard to review. Could you please rebase your branch to the current `master` branch of Wegue? I guess your commit [c7ce6d5](https://github.com/wegue-oss/wegue/pull/381/commits/c7ce6d5e98684bc6cd77aec6d37e861adfffd5f3) is the one of relevance. Thanks in advance. Afterwards I'll happily give it a review.

sronveaux:
Hi @enricofer and thanks for creating this PR!

I just took half an hour to have a look at the errors you encounter during tests as I was quite surprised that the dev and prod builds were working but not the test one... I unfortunately have not enough time to delve deeper into this for now but what could be done is compare the webpack configurations that are resolved in dev and test mode using, for example, `npx vue-cli-service inspect`.

In a first time, if you want to go further with your PR, you can just replace the following in the `vue.config.js` file:

```javascript
  transpileDependencies: [
    'jspdf'
  ]
```

With this you'll be able to go on further until the reason why `jspdf` is correctly transpiled in certain circumstances only is found.

By the way, I can't make your PR run smoothly for now as the `wgu-print-win` component is not registered properly. There is certainly something missing in the `app-starter/WguAppTemplate.vue` file...

Thanks again for your interest in this project,

All the best

PR #383, opened 2024-04-24
Allow switching base layer by keyboard (ENTER)

chrismayer:
This enables the switch of the map's base layer in the `BgLayerSwitcher` component by hitting the ENTER key instead of clicking only.


chrismayer:
Thanks for your review @sronveaux.

PR #384, opened 2024-04-24
Make OpenLayers keyboard navigation work

chrismayer:
Adds `tabIndex` attribute to the map's container, so it gets focusable. Otherwise the OL keyboard navigation won't work, see `keyboardEventTarget` in https://openlayers.org/en/latest/apidoc/module-ol_Map.html

### DIFF
--- a/src/components/ol/Map.vue
+++ b/src/components/ol/Map.vue
@@ -78,6 +78,11 @@ export default {
         me.map.updateSize();
       });
       resizeObserver.observe(container);
+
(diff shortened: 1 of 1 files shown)


sronveaux (approved):
This greatly improves accessibility in Wegue. Too bad this is not better documented in OpenLayers, you were lucky to come across this one ;-)

It works flawlessly and LGTM. Ready to merge !

chrismayer:
Thanks again for your review @sronveaux :+1:

PR #386, opened 2024-05-01
Wmts support

enricofer:
this PR includes support for WMTS layer


fschmenger (changes requested):
Hi @enricofer,
thanks a lot for providing this!
I have to admit I`m not the OL expert on the team but i decided to have a brief look:

Since the main branch has progressed quite a bit:

* Could you merge it again please?
I also noticed that you commited 8ae5d214b3409ef75f0182f63ef2d80002d61870 - which probably does not belong into here - and then got rid of it later on in merge. This is a bit confusing. If possible it would be preferred if you could just rebase this PR on the current tip.
* Also, could you replace fetch by axios please?

What I can see currently missing is:
* A documentation of your enhancement of the `TileGridDefs` in `docs/wegue-configuration.md`
* Some unit tests for the WMTS option in `tests/unit/specs/factory/Layer.spec.js`

I made some more comments inline.

From a structural point of view, I'm trying to make up my mind whether we should make the `createTileWmtsLayer` function async, since an incomplete layer may be returned until fetching resolves. ( This however would infect the whole LayerFactory ). We could also just document on this in the description or put a TODO there and resolve it later. @chrismayer @sronveaux Any ideas welcome here.

I`ll be off for the next 3 weeks, so take your time!

Cheers, Felix

PR #387, opened 2024-05-02
Refactor VueI18n creation to own module

chrismayer:
This refactors the creation of the `VueI18n` instance by encapsulating it in an own ES module. So the i18n instance can be imported to other pure ES modules and will allow internationalization there.

This follows the approach given here: https://github.com/kazupon/vue-i18n/issues/149#issuecomment-357455921

This would make the introduction of a global `$appLanguage` variable in #356  obsolete by this much cleaner and more versatile approach. If you approve this PR, I'd happily revert #356 afterwards in a separate PR.

Accessing the current app language in pure ES modules could then be done by the following code:

```js
const { i18n } = await import('../../../src/locales/wgu-i18n.js');
console.log(i18n.locale);
```

### DIFF
--- a/src/locales/wgu-i18n.js
+++ b/src/locales/wgu-i18n.js
@@ -1,0 +1,18 @@
+/**
+ * Creates the VueI18n object used for internationalization.
+ */
+import Vue from 'vue';
+import VueI18n from 'vue-i18n';
+import LocaleUtil from '@/util/Locale';
+
+Vue.use(VueI18n);
+
+const appConfig = Vue.prototype.$appConfig;
+
+const preset = {
+  locale: LocaleUtil.getPreferredLanguage(appConfig),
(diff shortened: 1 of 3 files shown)


sronveaux (approved):
Works completely as expected. As said, this one renders #356 obsolete.

The only thing I wanted to comment on is that assignment of `$appConfig` on `Vue.prototype` directly inside unit tests is considered bad practice. We should create a local Vue instance and assign to this local instance in place of the global one.  
However, as this is also done in other modules (for example `LocaleSwitcher.spec.js`), I approved as is. Perhaps an issue should be created with this point to be addressed and enhanced later...

chrismayer:
Thanks for your review @sronveaux and thanks for your suggestion regarding the assignment to `Vue.prototype` in the tests. I created an issue to improve that as a whole :arrow_right: #388. Going to merge now...

PR #389, opened 2024-05-15
Prepare workshop for upcoming v2.0.0

chrismayer:
This prepares the Wegue workshop for the upcoming version 2.0.0. Also corrects some minor issues.

I also went through the workshop and it is still functional with the new future version.

### DIFF
--- a/docs/workshop.md
+++ b/docs/workshop.md
@@ -1,6 +1,6 @@
 # Workshop
 
-This workshop uses Wegue version [`v1.2.1`](https://github.com/wegue-oss/wegue/releases/tag/v1.2.1) but works for higher versions with no or little adaptation.
+This workshop uses Wegue version [`v2.0.0`](https://github.com/wegue-oss/wegue/releases/tag/v2.0.0) but works for higher versions with no or little adaptation.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


sronveaux (approved):
Hi @chrismayer,

Good thing, you managed to spot things I missed when doing #361 and some changed texts are clearer now !

There is one thing that would deserve to be fixed in the workshop I think (but in another PR) as we finish the  "Use a module" part with a modified theme (pink borders and footer texts changed) then do the "Custom functionality" part as if we didn't make those changes and didn't add the layers and zoom-to-max-extent modules to eventually get to the "Add a custom module" part which resumes as if "Custom functionality" didn't exist...  
I tried to find how changes could be minized at that time and think we could address this by taking new screenshots for the "Custom functionality" part and change the beginning text of "Add a custom module" with something like `So far we modified the functionality of Wegue using the configuration file or by writing code into WguApp.vue. Now we will create a custom module using JavaScript. First we will revert what was made in WguApp.vue in the previous section. Then, to effectively create the custom module, we need to change a couple of files:`

This is only some thoughts about something that could be addressed later though... concerning this specific PR, it LGTM and is ready to merge !

chrismayer:
Thanks for your review and feedback @sronveaux :+1: I created an issue based on your comment. So we can tackle this later and will not be forgotten...

PR #396, opened 2024-06-04
Fix hoverable WMS layers.

fschmenger:
Hoverable WMS layers didn`t work because the respective properties were accidently forwarded to the source instead of the layer. It seems that this bug has been around ever since the first implementation.

### DIFF
--- a/src/factory/Layer.js
+++ b/src/factory/Layer.js
@@ -129,11 +129,11 @@ export const LayerFactory = {
         ratio: lConf.ratio,
         interpolate: lConf.interpolate,
         projection: lConf.projection,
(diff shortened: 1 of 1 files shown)


sronveaux (approved):
Wow, it is so obvious that it stayed buggy for such a long time...

I fully tried it, the hardest part for me was to find a public WMS server that replied with GML... I must admit I never used some beforehand...

I can confirm that everything works as expected !  This should be ready to merge !

chrismayer (approved):
Nice catch @fschmenger :+1: 

I tested and it works now as expected. Thank you very much.

PR #397, opened 2024-06-06
Unregister ResizeObserver for map container

chrismayer:
This unregisters the `ResizeObserver` for the map's container element, when the map is unmounted/destroyed.

Besides having a better cleanup this also removes the null pointer access within the `ResizeObserver` handler in the hot-reload of the dev- setup, when then `WguApp.vue` is changed.

### DIFF
--- a/src/components/ol/Map.vue
+++ b/src/components/ol/Map.vue
@@ -78,6 +78,7 @@ export default {
         me.map.updateSize();
       });
       resizeObserver.observe(container);
+      this.resizeObserver = resizeObserver;
 
       // add tabIndex attribute to the map's container, so it gets focusable.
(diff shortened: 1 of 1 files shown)


fschmenger (approved):
Nice catch, I have currently no time to test but totally makes sense.

For paranoia reasons I suggest to wrap up the `disconnect` with an
`if (this.resizeObserver)`
just in case `container` is undefined on registration.

chrismayer:
Thanks for your review @fschmenger!

I added the check `if (this.resizeObserver)`, not just for paranoia :wink: - it was necessary to make the tests green. Here we exactly have the situation, that no container exists. Thanks for the hint!

PR #399, opened 2024-06-06
Remove non-effective color prop in ToggleButton

chrismayer:
This removes non-effective color prop definition in the `ToggleButton` component.

### DIFF
--- a/src/components/modulecore/ToggleButton.vue
+++ b/src/components/modulecore/ToggleButton.vue
@@ -7,7 +7,7 @@
     class="wgu-toggle-button"
     :title="$t(moduleName + '.title')"
     v-model="show">
-    <v-btn icon :value="true" color="onprimary" @click="toggleUi">
+    <v-btn icon :value="true" @click="toggleUi">
       <v-icon color="onprimary" medium>{{icon}}</v-icon>
     </v-btn>
   </v-btn-toggle>


src/components/modulecore/ToggleButton.vue
```diff
@@ -2,12 +2,11 @@
   <v-btn-toggle
     borderless
     dense
-    color="onprimary"
```
sronveaux, line 5: Are you sure you want to remove this line ?

Here's what it looks with it : 

image

Here's what it looks without :

image

It's a matter of taste for sure but PR title says remove `non-effective` color props and this one is...

src/components/modulecore/ToggleButton.vue
```diff
@@ -2,12 +2,11 @@
...
     background-color="transparent"
     class="wgu-toggle-button"
     :title="$t(moduleName + '.title')"
     v-model="show">
-    <v-btn icon :value="true" color="onprimary" @click="toggleUi">
+    <v-btn icon :value="true" @click="toggleUi">
```
sronveaux, line 9: This one is effectively useless !

sronveaux (commented):
Hi @chrismayer,

Just made the test and have to say that only one of the rules you removed are truly non-effective.

I prefer to let you see whether this should be amended or whether it should stay the same... but in this case, PR name should be changed as one of the removed lines IS still effective...

fschmenger:
Maybe add @ntma as a reviewer and wait until he is around?
A lot of work and testing went into theming, so rather be sure there are no unintended side effects.

src/components/modulecore/ToggleButton.vue
chrismayer: Totally correct, I did not recognize that. Either I happened to have a setup with the same colors or my old eyes fooled me. Glad you caught that! I'll come up with a correction.

sronveaux (approved):
Looks perfect from my side and is ready to merge !

I let you appraise whether there is a chance to get @ntma comment on this... I suppose this can be changed again in the future in that case !

chrismayer:
Thanks for your additional review @sronveaux. I am going to merge now, since I don't expect any problems by this changeset. @ntma If I am wrong, please comment here (even after merging) and I'll happily come up with separate PR.

PR #403, opened 2024-06-14
Set version to 2.0.1

chrismayer:
This sets the version tags in the code to 2.0.1 as a final preparation step for the upcoming 2.0.1 release.


chrismayer:
Merging without review since this is only a formal (non-functional) step within the release process.

PR #404, opened 2024-06-18
Allow custom theme properties

chrismayer:
This allows to define custom theme properties in the app-config JSON, like

```json
  "colorTheme": {
    "dark": false,
    "themes": {
      "light": {
        "primary": "#FFFFFF",
        "onprimary": "#000000",
        "secondary": "#B9B9B9",
        "cmcustom": "#000074"
      },
      "dark": {
        "primary": "#FF0000",
        "secondary": "#FFCC80",
        "onprimary": "#00FF00",
        "cmcustom": "#FF0074"
      }
    }
  },
```
Before the theme properties were limited to the default ones of Vuetify. According to[ Vuetify2 theme docs](https://v2.vuetifyjs.com/en/features/theme/#custom-theme-variants) it is totally OK to define custom properties and very useful IMHO. The custom properties are fully treated as the default ones and the derivatives like `-lighten1` or `-darken5` will be created.

### DIFF
--- a/docs/wegue-configuration.md
+++ b/docs/wegue-configuration.md
@@ -32,7 +32,7 @@ This describes the Wegue application configuration, which is modelled as JSON do
 
 ### colorTheme
 
-Wegue supports the Vuetify (light and dark) theme configuration out of the box. This property expects a json object with the same format described in the [vuetify documentation](https://vuetifyjs.com/en/features/theme/#customizing).
+Wegue supports the Vuetify (light and dark) theme configuration out of the box. This property expects a json object with the same format described in the [vuetify documentation](https://v2.vuetifyjs.com/en/features/theme/#customizing).
 
 Example:
 ```json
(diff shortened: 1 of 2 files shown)


sronveaux (approved):
Tested and works as explained and expected !

I was just wondering whether this would deserve a line in the documentation or not... something like :

In addition, Wegue also supports the following "on" colors:

| Color         | Description         | Example       |
| ------------- |:-------------------:| ------------- |
| onprimary     | color over primary color   | typography/icons over primary color   |
| onsecondary   | color over secondary color | typography/icons over secondary color |

**Moreover, custom colors which are not listed in the Vuetify documentation can also be defined here. Lighten and darken variants will also be generated for them like for the default ones.**

To simplify the theming configuration, ...

Final remark here, the link to Vuetify documentation in `wegue-configuration.md` is currently not the good one as it links to Vuetify3 documentation which has some differences... should it be amended to https://v2.vuetifyjs.com/en/features/theme/#customizing (and reverted when Wegue will be ported to Vue3) or not... I let you choose the best option...

Other than that, on a functional point of view, it's perfect and ready to merge !

chrismayer:
Thanks for your feedback @sronveaux. I updated the theming documentation with your suggestions.

Going to merge now. Failing tests seem to be unrelated, since they were green before without any substantial change and they also fail for #405. Investigation should be done separately...

PR #405, opened 2024-06-20
Unregister events properly inside AttributeTable component

sronveaux:
This PR cleanly unregisters the two events that are used inside the AttributeTable component.

One event is called when a feature is selected on the map. This was still registered after the component was closed...

The second is called when the 'onchange' event of the selected layer is fired. This was not cleanly unregistered when changing layers, so selecting a feature of the previously selected layer continued to fire this event and stayed even after the component was closed.

Should be fixed now...


sronveaux:
Tests now pass thanks to @chrismayer's celerity with #406 !

PR #406, opened 2024-06-20
Fix MapRecorderWin unit tests

chrismayer:
This fixes the unit tests for the `MapRecorderWin`. Reason is that there is an addition in the mimeType list of the browser. Additionally `video/mp4` is listed there now.

### DIFF
--- a/tests/unit/specs/components/maprecorder/MapRecorderWin.spec.js
+++ b/tests/unit/specs/components/maprecorder/MapRecorderWin.spec.js
@@ -47,7 +47,7 @@ describe('maprecorder/MapRecorderWin.vue', () => {
       expect(vm.error).to.equal(false);
       // Supported codecs under chrome.
       expect(vm.mimeType).to.equal('video/webm');
-      expect(vm.mimeTypes.length).to.equal(2);
+      expect(vm.mimeTypes.length).to.equal(3);
     });
 
     afterEach(() => {
@@ -65,9 +65,10 @@ describe('maprecorder/MapRecorderWin.vue', () => {
 
     it('correct supported mime types under chrome', () => {
       const mimeTypes = vm.getSupportedMimeTypes();
-      expect(mimeTypes.length).to.equal(2);
+      expect(mimeTypes.length).to.equal(3);
       expect(mimeTypes[0]).to.equal('video/webm');
-      expect(mimeTypes[1]).to.equal('video/x-matroska');
+      expect(mimeTypes[1]).to.equal('video/mp4');
+      expect(mimeTypes[2]).to.equal('video/x-matroska');
     });
 
     it('start / stop recording sets correct states', () => {


tests/unit/specs/components/maprecorder/MapRecorderWin.spec.js
```diff
@@ -47,7 +47,7 @@ describe('maprecorder/MapRecorderWin.vue', () => {
       expect(vm.error).to.equal(false);
       // Supported codecs under chrome.
       expect(vm.mimeType).to.equal('video/webm');
-      expect(vm.mimeTypes.length).to.equal(2);
+      expect(vm.mimeTypes.length).to.equal(3);
```
sronveaux, line 50: I was wondering whether this line should stay or be completely removed... the test is about default values... from a point of view, it is assigned when building the data object but on the other, it's value is computed and dependent of the browser the test is running on... so is it really a default value ?

I let you take the decision...

```suggestion
```

tests/unit/specs/components/maprecorder/MapRecorderWin.spec.js
```diff
@@ -65,9 +65,11 @@ describe('maprecorder/MapRecorderWin.vue', () => {
 
     it('correct supported mime types under chrome', () => {
       const mimeTypes = vm.getSupportedMimeTypes();
-      expect(mimeTypes.length).to.equal(2);
+      console.log('MIM', mimeTypes);
```
sronveaux, line 68: Debugging trace left here ?

```suggestion
```

sronveaux (approved):
Great !  As you stated, this change has started to make all test runs to fail and I was also going to post a PR to correct this as I encountered the same problem in #405...

I was just thinking whether the test on default Chrome media possibilities was really useful written as is... it is run whatever the browser which is used to run the tests and so doesn't always give the expected result... and even with Chrome, it is only relevant if using the correct version of the browser... Perhaps it would be good to open a separate issue to discuss this and see how to address it... I'd personally write a test to see if at least one format was available on the browser instead of testing a fixed list but it's only a suggestion... I can do it if everybody agrees on it...

For now, this PR fixes the reported problem and should be merged for sure !

Thanks @chrismayer to have done it so fast !

tests/unit/specs/components/maprecorder/MapRecorderWin.spec.js
chrismayer: Sure :see_no_evil: , I'll correct it.

chrismayer:
> I was just thinking whether the test on default Chrome media possibilities was really useful written as is

I was thinking the same when analyzing the code for the fix. I#d gladly want to discuss this with you and the others. So I created an issue #407 for that, which will be the base for a possible adaption of the test.

Goal of this PR was to get the test suite green again. So, I am going to merge now.

Issue #116, opened 2020-03-18
Wegue project logo


chrismayer:
Having a project logo would be very nice. 

One very rough first idea: since the name is pronounced like the Japanese cattle breed "Wagyū" it would be cool if the logo contains a cow / bull / cattle / buffalo / etc. And maybe something what gets the connect to the "geo" world.

chrismayer:
:laughing: :laughing: :laughing: :laughing: :laughing: just kidding

![grafik](https://user-images.githubusercontent.com/1185547/76945954-95f68180-6903-11ea-941d-f993517496f8.png)

justb4:
We have a GeoMoose already as a large herbivore, so ok in that line. On DuckDuckGo "Wagyū" searches refer a lot to its meat like on Austrian sites: https://wagyu-oesterreich.at/wagyu-rind/ . In these times the project may not want to be associated with beef cattle used for consumption, but this is my personal opinion here. Though a lot of nice open source projects refer/are named by or after cows (old days: https://archive.org/details/tucows) , cattle and herding (e.g. Rancher), I had long ago project CowCatcher: http://justobjects.org/cowcatcher/. ...And GNU, but that is a wild buffalo...

I was already wondering how to pronounce "Wegue" but from the tones of "Web" and "Vue" then indeed it becomes like "Wagyū". Many more associations are possible. "yū" could even refer to Evan You, creator of Vue.js. "Wag" to tail-wagging. But then again Wag refers to this: https://en.wikipedia.org/wiki/WAGs In short: open for anything...

JakobMiksch:
I asked my sister about creative ideas, she came up with something like this, but it is in an early stage :-)

![image](https://user-images.githubusercontent.com/15704517/77193693-f392f580-6ade-11ea-8edb-5585ae9d52f8.png)

chrismayer:
I did 2 rough sketches for a logo:

![image](https://user-images.githubusercontent.com/1185547/107626634-627cbe00-6c5e-11eb-93aa-0d3e85b439f5.png)

![image](https://user-images.githubusercontent.com/1185547/107626741-86400400-6c5e-11eb-8233-4a62cc74929c.png)

I like the first one better. Opinions?

JakobMiksch:
Looks good so far. 
We should make sure that it is still visible when it is small  e.g. as a  favicon or so ...

justb4:
Suggestions: keep the logo more simple. It has to be used in many contexts and sizes. 
[...]

PR #409, opened 2024-07-12
Allow Vuetify custom icons

sronveaux:
This PR opens the possibility to define custom icons to be injected inside Vuetify.

screenshot

Just to give a little bit more context and explain some design choices around this:

Three methods were tested to define and import those custom icons:

1. Defining an SVG path exported as a string from a `.js` file. This method which is the one implemented in this PR is the simplest one which mimicks how mdi itself is written ([see](https://github.com/Templarian/MaterialDesign-JS/blob/master/mdi.js)).
2. Importing SVG files directly using an appropriate loader. This methods necessitates to add a custom loader for `Webpack` and tweak the configuration. The standard loaders available out there are often not up-to-date and don't import icons in a format which can be resized or colorized properly as [explained in Vuetify documentation](https://v2.vuetifyjs.com/en/features/icon-fonts/#component-icons). A way to circumvent this would be to write our own `Webpack` loader as a separate package to publish and maintain on NPM which is a bit overkill from my point of view.
3. Importing VUE components which embeds an SVG directly (see [the same page as for point 2](https://v2.vuetifyjs.com/en/features/icon-fonts/#component-icons)) which works but necessitates to rewrite a component for each and every icon and to add styling inside for it to embed properly.

A first implementation was made that imported those three formats, which explains the way the `IconUtil` file is written. This structure was kept in case such a behaviour was needed in the future.

However, only the first method was kept in the end as this is the only one that can be migrated to `Vuetify3` easily. Components would need to be written very differently to be used correctly by `Vuetify3` which makes the second and third method useless for now  as someone writing icons in such formats would need to rewrite them completely when migrating later...

If this PR is eventually accepted and merged, just let me know if you want me to add this to the workshop. We could for example demonstrate how to change the icon of the custom component. This could be great and funny if a decision is taken about #116 one day !

### DIFF
--- a/src/util/Icon.js
+++ b/src/util/Icon.js
@@ -6,9 +6,9 @@ import ObjectUtil from './Object.js';
 const IconUtil = {
 
   /**
-   * Import custom icons from 'app/icons'. Those icons will be included in an
-   * object ready to be injected inside Vuetify configuration options object under
-   * the icons['values'] property.
+   * Import custom icons from 'app/custom-icons'. Those icons will be included
+   * in an object ready to be injected inside Vuetify configuration options
+   * object under the icons['values'] property.
    *
    * @returns {Object} An object containing the imported icons.
    *                   Key is the icon name, value contains the icon.
@@ -20,15 +20,17 @@ const IconUtil = {
       const pathIcons = IconUtil.importPathIcons();
       ObjectUtil.mergeDeep(icons, pathIcons);
     } catch (e) {
+      console.warn('An error occured while importing the custom icons, ' +
+        'they will not be registered inside Vuetify. The error was: ', e);
     }
 
     return icons;
   },
 
   /**
-   * Import custom icons from 'app/icons/*.js'. Those icons will be included in an
-   * object ready to be injected inside Vuetify configuration options object under
-   * the icons['values'] property.
+   * Import custom icons from 'app/custom-icons/*.js'. Those icons will be included
+   * in an object ready to be injected inside Vuetify configuration options object
+   * under the icons['values'] property.
    * The icons must be .js files with a default export which contains an SVG path.
    *
    * @returns {Object} An object containing icon SVG paths.
@@ -36,12 +38,12 @@ const IconUtil = {
    */
   importPathIcons () {
     const moduleDefaultExtractor = (i) => i.default;
-    const testExp = /icons\/(?:.+\/)*([a-z0-9_-]+).js$/i;
+    const testExp = /custom-icons\/(?:.+\/)*([a-z0-9_-]+).js$/i;
 
     const context = require.context(
       '../../app',
       true,
-      /icons\/(?:(?:.+)\/)*([a-z0-9_-]+).js$/i
+      /custom-icons\/(?:.+\/)*([a-z0-9_-]+).js$/i
     );
     const pathIcons = {};
     for (const key of context.keys()) {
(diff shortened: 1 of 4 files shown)


src/util/Icon.js
```diff
@@ -0,0 +1,56 @@
...
+   * @returns {Object} An object containing the imported icons.
+   *                   Key is the icon name, value contains the icon.
+   */
+  importIcons () {
+    const icons = {};
+
+    try {
+      const pathIcons = IconUtil.importPathIcons();
+      ObjectUtil.mergeDeep(icons, pathIcons);
+    } catch (e) {
```
chrismayer, line 22: Maybe add a warning for developers in the `catch` block?

chrismayer (approved):
Thanks for this nice feature @sronveaux :+1: I like the possibility to add custom SVG icons and the simplicity. I second your choice not to go for a custom Webpack loader et.al. The approach is easily portable to future developments.

Code looks good and I tested it successfully. 

One little remark, since I mixed it up while testing: There is already a folder `icon` and I placed my custom SVG-icon in there and it surely did not show in the app since it is the wrong folder.  Your extension looks in folder `icons`. Maybe another name could reduce the chance to mess those folders up, such as `custom-icons` or similar? Or is it just my dumbness? I leave it up to you, whether to change or not. I am fine with both.

I also like the idea to have this in feature in the workshop. A perfect spot to illustrate the possibilities.

sronveaux:
Hi @chrismayer and thanks for the review !

I'll address the comment you made shortly...

Just coming back to you to have your opinion about the `icons` folder. Sure I can rename it without problem. As stated in the docs, it is not in `app/static` where you already find the `icon` folder but directly in `app/icons`, as placing them inside `static` would also copy them in the `dist` folder while building. As they need to be bundled to get caught by `Vuetify`, I got them out of `static`.

I also thought about creating an empty `icons` directory inside the `app-starter` so people would directly see where to put them. That's also why I had to "overcomplicate" the regex looking for icons as the `icons` directory doesn't necessarily exist... Just wanted to know if you would find it clearer if there was an empty `custom-icons` created inside the `app-stater` ?

And thinking about the future, do you have an idea for the icon to assign in the workshop ?  The `W` I've put in the docs looks a little too simple I think... Perhaps I can reuse [the idea of @justb4](https://user-images.githubusercontent.com/582630/109134764-af7c8c00-7756-11eb-941f-bd234cff6bc6.png) or do you have something else in mind ?

src/util/Icon.js
sronveaux: Warning was added as suggested

sronveaux:
Hi @chrismayer,

I should have addressed all you have suggested here... could you have another look please ?

chrismayer:
Very nice, thanks for your ongoing effort here @sronveaux! Looks good from my side. Please merge at will :+1:

chrismayer:
> And thinking about the future, do you have an idea for the icon to assign in the workshop ? The W I've put in the docs looks a little too simple I think... Perhaps I can reuse [the idea of @justb4](https://user-images.githubusercontent.com/582630/109134764-af7c8c00-7756-11eb-941f-bd234cff6bc6.png) or do you have something else in mind ?

With some temporal distance I have to admit that I like the logo suggestion you linked. If you are familiar with graphics or you want to present/discuss something let's move the discussion in the existing issue #116 Would be really cool to have a logo after all the time...

PR #410, opened 2024-07-19
Fix Photon geocoder provider

sronveaux:
This PR fixes the Photon geocoder provider. There were two problems with it :

1. The URL has changed four years ago, even though the old one still works, it is a good idea to update it as stated on [photon homepage](https://photon.komoot.io/)
2. Photon returns the bounding boxes with Y coordinates flipped which breaks the move to a selected place. A test was added to put the coordinates in the [format expected by OpenLayers](https://openlayers.org/en/latest/apidoc/module-ol_extent.html#~Extent)

### DIFF
--- a/src/components/geocoder/providers/photon.js
+++ b/src/components/geocoder/providers/photon.js
@@ -23,7 +23,7 @@
 //
 
 /**
- * Photon Provider. See https://photon.komoot.de.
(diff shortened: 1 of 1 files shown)


chrismayer (approved):
Cool, thanks for tackling this. Looks good to me. Please merge at will...

PR #415, opened 2024-09-11
French translation of Wegue

sronveaux:
Here is the french translation of Wegue I promised a few months ago now...


chrismayer (approved):
Thanks for this, very much appreciated :+1: 

Technically good to go and I tested the example application. No review for the texts itself since my French is more than rusty :wink: 

Please merge at will @sronveaux

PR #416, opened 2024-09-20
Geocoder refactoring using axios and AbortController objects. Fixes #400

### LINKED ISSUE #400: Use axios consequently for all HTTP requests in Wegue
Since `axios` is added to Wegue as dependency we should consequently use `axios` to perform HTTP requests. So usage of other techniques like `fetch`, `XMLHttpRequest`, ... should be replaced by `axios` calls.

Herewith axios should be upgraded to its latest version.

This issue is based on this [discussion](https://github.com/wegue-oss/wegue/issues/394).

sronveaux:
This PR is a refactoring of the Geocoder using axios and AbortController to... abort requests when user asks for something else.

Globally, it makes the component more reactive. Before, if you typed an address while a request was running, the new address would not get searched. It should now be correctly debounced.

This PR also removes `JSONP` which was not used anymore.

It fixes #400 as axios will then be used in all requests made inside `Wegue`.

Eventually, unit tests are refactored too to use `axios-mock-adapter` on one side and to use `sinon` timers on the other.  
Looks like `sinon` timers are now working well from my side, but please give it a try to ensure it works as expected on other configurations too...


chrismayer (approved):
Thanks for this. Tested it and it works like charm. Please merge at will...

PR #419, opened 2024-09-25
Corrected InfoClick bug when using MeasureTool afterwards

### LINKED ISSUE #285: InfoClickWin: Unregistering of listeners not working
The listeners of the `InfoClickWin` are not unregistered. See here: https://github.com/meggsimum/wegue/blob/master/src/components/infoclick/InfoClickWin.vue#L146-L154
I can probably fixed with:

```js
 if (visible) {
        me.registerMapClick();
      } else {
        me.registerMapClick(true); // <-- THIS IS THE FIX
    [...]
}
```

Normally, this bug is not noticed, but can have effects in special situations.

sronveaux:
Since addition of #367, after `InfoClick` was opened, using some other tools like `MeasureTool` would break at first click on the map.  
This is linked to the fact that the geometry drawn by the measure tool (points and lines) are returned by `forEachFeatureAtPixel`. However, those aren't linked to any layer which made `viewProps` fail in this case.

To correct it, `features` which are not assigned to a layer are not kept by `forEachFeatureAtPixel`.

This PR also corrects another linked bug where the map click event was not unregistered properly when `InfoClick` component was closed. This was apparently reported some time ago. So this also fixes #285

### DIFF
--- a/src/components/infoclick/InfoClickWin.vue
+++ b/src/components/infoclick/InfoClickWin.vue
@@ -128,7 +128,9 @@ export default {
       me.features = []
       me.map.forEachFeatureAtPixel(evt.pixel,
         (feature, layer) => {
-          me.features.push([feature, layer])
+          if (layer) {
+            me.features.push([feature, layer])
+          }
         });
 
       // collect feature attributes --> PropertyTable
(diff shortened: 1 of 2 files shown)


chrismayer (approved):
Nice catch @sronveaux! Thanks for tackling these 2 problems at once. 

Tested and reviewed :arrow_right:  LGTM :arrow_right: please merge at will :+1:

PR #420, opened 2024-10-02
Upgrade OpenLayers to 10.2.1

chrismayer:
This upgrades the OpenLayers dependency to version 10.2.1. No special actions are needed in the Wegue code base, since it is only a bugfix release. See official [OL changelog](https://github.com/openlayers/openlayers/blob/main/changelog/v10.2.1.md).


sronveaux (approved):
Just tested, as always, on a bigger app and can confirm all work as expected without any observable flaw.

Ready to merge!

PR #426, opened 2025-03-13
Set version to '3.0.0-dev' in package(-lock).json

chrismayer:
Sets the version property to `3.0.0-dev` in `package.json` and `package-lock.json` to clearly transport that the master branch is used for development of upcoming Wegue v3.


sronveaux (approved):
That's perfect and a good idea to do this now...

Approved and ready to merge!

PR #427, opened 2025-03-25
Fix featuresloadend and featuresloaderror events not fired properly by WFS layers

sronveaux:
There is currently a bug where `featuresloadend` and `featuresloaderror` events are not fired properly by WFS layers.

This is due to missing calls to callback functions inside the underlying `vectorSource` as described inside the [official OpenLayers documentation](https://openlayers.org/en/latest/apidoc/module-ol_source_Vector-VectorSource.html) inside the definition of the `loader` constructor option:

_The `featuresloadend` and `featuresloaderror` events will only fire if the success and failure callbacks are used._

This PR also refactors the WFS layer factory function to use Axios properly. This could have been a separated PR but was included here as it was a mandatory step to make the events firing unit tests writeable...

### DIFF
--- a/app-starter/static/app-conf-projected.json
+++ b/app-starter/static/app-conf-projected.json
@@ -78,7 +78,7 @@
     {
       "type": "WFS",
       "lid": "dutch-nat-parks",
-      "url": "https://service.pdok.nl/rvo/nationaleparken/wfs/v2_0?",
+      "url": "https://service.pdok.nl/rvo/nationaleparken/wfs/v2_0",
       "typeName": "nationaleparken:nationaleparken",
       "version": "2.0.0",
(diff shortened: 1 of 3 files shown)


chrismayer (approved):
Very nice, and good catch @sronveaux :fishing_pole_and_fish: 

Code looks good and I tested, WFS layers work properly. Feel free to merge. Thank you!

chrismayer:
We should consider back-porting this fix to the [v2](https://github.com/wegue-oss/wegue/tree/v2) branch.

sronveaux:
Hi @chrismayer,

Very good idea to backport this for sure !

Could you please give some information or send me an e-mail with details about the workflow and/or strategy you'd like to apply when backporting ?

PR #428, opened 2025-03-25
Refactoring of Map Loading Status component

sronveaux:
Current implementation of Map Loading Status component has some problems and doesn't detect loading with all and every type of layers.

This is mainly due to the fact that it was written at the time of `OpenLayers 5.3.3` which was released in April 2019 and wasn't upgraded afterwards.

Since `OpenLayers 6.14` (released in March 2022) two map events have been added, `loadstart` and `loadend` which greatly simplifies the job to be done and should be compatible with all upcoming layer types that could be added later.

You can refer to [ahocevar PR](https://github.com/openlayers/openlayers/pull/13491) talking about that.

This PR refactors the Map Loading Status component, using those two events which permits to have a stable behaviour whatever the source and layer types added to the map.

Unit tests are also included for the component.

### DIFF
--- a/src/components/progress/MapLoadingStatus.vue
+++ b/src/components/progress/MapLoadingStatus.vue
@@ -11,11 +11,6 @@
 
 <script>
 import { useMap } from '@/composables/Map';
-import {
-  Image as ImageSource, TileImage as TileImageSource,
-  Vector as VectorSource, Cluster as ClusterSource
-} from 'ol/source';
-import LayerGroup from 'ol/layer/Group';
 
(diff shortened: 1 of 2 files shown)


chrismayer (approved):
Very cool, this simplifies a lot :muscle:  Please merge at will. Thank you.

chrismayer:
Should we back-port this to [v2](https://github.com/wegue-oss/wegue/tree/v2) or is this rather an improvement/refactoring than a fix? What do you think @sronveaux ?

PR #434, opened 2025-04-02
Ensure "SLD_VERSION" is set in "GetLegendGraphic" for WMS 1.3.0

chrismayer:
The PR ensures that the mandatory parameter `SLD_VERSION` is applied to the `GetLegendGraphic` requests for WMS 1.3.0, if not set already.

### DIFF
--- a/src/util/Object.js
+++ b/src/util/Object.js
@@ -55,6 +55,34 @@ const ObjectUtil = {
       result.push(path);
     }
     return result;
+  },
+
+  /**
+   * Returns the value from an object using a case-insensitive key.
+   *
+   * @param {Object} obj The object to search in
+   * @param {*} key The key to retrieve value for
+   * @returns {*} The value for the given key or undefined, if not found
+   */
+  getValueIgnoreCase (obj, key) {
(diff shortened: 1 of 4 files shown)


src/util/Object.js
```diff
@@ -55,6 +55,34 @@ const ObjectUtil = {
       result.push(path);
     }
     return result;
+  },
+
+  /**
+   * Returns the value from an object using a case-insensitive key.
+   *
+   * @param {Object} obj The object to search in
+   * @param {*} key The key to retrieve value for
```
sronveaux, line 64:
```suggestion
   * @param {String} key The key to retrieve value for
```

Shouldn't key be declared as String so IDEs can report wrong usage ?

sronveaux (approved):
Thanks for this catch and nice additions !

Tested and it works like a charm !

I just added a change suggestion but let you see if it should be taken into account or not...

Already approved the PR by the way !

sronveaux:
Thanks for amending @chrismayer, LGTM and ready to be merged !

chrismayer:
Thanks for your review @sronveaux. I am going to merge now.

PR #437, opened 2025-04-02
Backport #428: Refactoring of Map Loading Status component

### LINKED ISSUE #430: Backport v2 #428: Map Loading Status component doesn't work properly with all type of layers
Current implementation of `Map Loading Status` component has some problems and doesn't detect loading with all and every type of layers.

This is mainly due to the fact that it was written at the time of `OpenLayers 5.3.3` which was released in April 2019 and wasn't upgraded afterwards.

Since `OpenLayers 6.14` (released in March 2022) two map events have been added, `loadstart` and `loadend` which greatly simplifies the job to be done and should be compatible with all upcoming layer types that could be added later.

This issue was already fixed in the master branch by #428 but should also be backported on the v2 branch.

sronveaux:
This PR backports #428 fix on the `v2` branch.

Fixes #430

### DIFF
--- a/src/components/progress/MapLoadingStatus.vue
+++ b/src/components/progress/MapLoadingStatus.vue
@@ -12,20 +12,13 @@
 </template>
 
 <script>
-
 import { Mapable } from '../../mixins/Mapable';
-import {
-  Image as ImageSource, TileImage as TileImageSource,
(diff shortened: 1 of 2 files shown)


chrismayer (approved):
Looks good to me, thank you also for backporting this :+1:

chrismayer (approved):

Issue #439, opened 2025-04-09
Geocoder Combox in WeGue3 is no longer able to adjust its size


bhubergeomer:
The Geocoder combobox in Wegue2 was able to adjust its size according to the available space in the toolbor.

In WeGue3 the width is hardcoded to 320.

Is it possible to reestablish the dynamic width?

chrismayer:
Hi @bhubergeomer, thanks for reporting this. I can confirm the described behavior. 

Indeed it would be better to have something like a [min-width](https://next.vuetifyjs.com/en/api/v-combobox/#props-min-width) on the `v-combobox` component level. Then you could adjust the width by CSS with

```css
.wgu-geocoder-combo {
  width: 500px;
}
```

which is not possible at the moment.

Would you be willing to bring up a Pull Request for this @bhubergeomer ? The change should really be easy by changing [this line](https://github.com/wegue-oss/wegue/blob/master/src/components/geocoder/Geocoder.vue#L8) to  `min-width="320"`. Would be very appreciated.

bhubergeomer:
Hi @chrismayer !

If i change to min-width the combobox looks fine and shows with the size of min-width, but it is not able to dynamically set its size according to the space available.
The Wegue2 version had a "flex" behavior, it gets smaller if there is little space.

In Wegue3 it was either the with of width paramter or of min-width parameter. It does not adjust its size. Some elements  of the Combobox DOM have display: grid. 
May this be a hint?

sronveaux:
Hi @chrismayer, hi @bhubergeomer,

Well, this one is more complicated than it sounds at first as behaviour in `Vuetify2` and `Vuetify3` are really different with this one. I'll try to summarise what can be observed:

In `Vuetify2`, the combobox had some kind of flex behaviour but with some limits as can be seen in the following screenshots taken fullscreen and with the screen diminished in size as much as possible:

![Image](https://github.com/user-attachments/assets/7cbdcdb6-a47a-4c9b-9dff-e7cd146379ee)

![Image](https://github.com/user-attachments/assets/32561132-c2ec-4560-80c0-a768d5bcbc6c)

Moreover, if you type even a single character inside the combobox, this size is changed:

![Image](https://github.com/user-attachments/assets/d5ef2f0c-2d75-4a8b-a598-796b687a2d1a)

![Image](https://github.com/user-attachments/assets/ae4606c8-482d-48bc-b366-59ac524fe36d)

[...]

PR #440, opened 2025-05-05
Fix header title getting truncated and change geocoder flex width behaviour

sronveaux:
This PR was written first to try and address #439 giving to the `Geocoder` combobox a min and max width which should give it a behaviour close to what could be observed in `Vuetify2` / `Wegue v1.x`. Those values can be adapted though.

At the same time, a bug where the header title gets truncated even when there is space left in the toolbar is addressed. This was [reported to Vuetify some time ago](https://github.com/vuetifyjs/vuetify/issues/18825). Answer was only given more than a year later in [another similar bug report](https://github.com/vuetifyjs/vuetify/issues/20966).


chrismayer:
Thanks for this @sronveaux. I think removing `<v-spacer></v-spacer>` is a must since this is a fix. 

Handling width, min-width, flex, etc. can become tricky as the discussion in #439 shows. I'd vote as a minimum to allow to configure `width` and `min-width` as `props`, so they could be set optionally in the module configuration. 

```css
:width="width"
:min-width="minWidth"
```
Without setting them the UI also looks quite nice IMHO: 

image

So this would be a minimal approach by exposing Vuetify properties.  Anything further regarding to a flex like behavior I am unsure, if we should implement that on Wegue level. But this we could discuss then in #439.

sronveaux:
Hi @chrismayer and thanks for your comment,

I agree with you except I'd expose `width`, `min-width` and `max-width` personally.

I also think playing with the flex behaviour can become 'dangerous' as it could conflict with `Vuetify` internal workings. That's the reason why I posted a hack that I was not in favour of including in the PR in the end. But if this helps someone...

Last question is which values should be set as default. I just did a quick test and saw that, opposite to `CSS` rules, `Vuetify` **currently** makes `width` take precedence over `min-width` and `max-width`.  
For sure they were set in an empiric manner but the values for min and max I've selected for this PR are the ones which closely match `Vuetify2` behaviour. I would definitely use them as default for this reason and that way, people would either have the opportunity to define their own min and max or fixed-width value.

If you agree with that, I'll amend the PR in the following days. Just tell me what you think about it...

chrismayer:
> I agree with you except I'd expose width, min-width and max-width personally.

So you mean, you would **not** expose them? So what would be the recommended way for devs to change the width of the UI? Via CSS?

sronveaux:
No no no !  I would definitely expose them !  But instead of exposing two properties (`width` and `min-width`) I would also add `max-width` !

chrismayer:
> No no no ! I would definitely expose them ! But instead of exposing two properties (`width` and `min-width`) I would also add `max-width` !

Oh perfect, then we are on the same page. I am also fine with setting defaults to come close to old v2 behavior. Thanks in advance for your input.

sronveaux:
O.K., PR is amended following preceding discussion.

There is a difference though, the fast tests I did had something wrong, in the end (and it's better like that !) `Vuetify` follows official `CSS` rules and `width` is superseded by `min-width` and `max-width`. However, the PR implements what was discussed:

1. Default behaviour is to have `minWidth` and `maxWidth` defined by default with values which closely match what was observed with `Vuetify2`.
2. The user can override those defaults in the configuration file if he wants.
3. The user can also set a fixed `width` value which takes precedence over `minWidth` and `maxWidth`.

Documentation is amended and unit tests have been added too.

chrismayer:
Thanks for your ongoing work on this @sronveaux. I am +1 to merge this. If we want to to further adaptations they should be done in separate PRs.

chrismayer (approved):

sronveaux:
Thanks @chrismayer, exactly what I wrote in #439. Let's see where the discussion will lead and whether we'll go further...  
I've given a proposal there even though I'm not in favor of implementing it but I'll follow what happens in this thread and reply if requested...

PR #441, opened 2025-07-21
Whs vuln detec 1

pxxguin:


justb4:
I think the regular expressions for the Path Traversal blocking and the icon naming are ok. Though I cannot see a direct vulnerability on the original code. The regex for icon naming is just more compact using `w`. 

But why @pxxguin did you add a [new GH Workflow file](.github/workflows/pr_notify.yml) that calls back to a Slack and a Discord channel?   AFAIK the project does not use these channels or I missed something. If you remove it we can re-review your PR.

pxxguin:
ok, thanks for your advice.

pxxguin:
I understood. I will re PR without pr_notify.yml!

PR #443, opened 2025-09-04
Maintenance v2.1.0

fschmenger:
This branch contains a couple of bugfixes, mostly related to the recent Vue3 migration. For details please see the commit messages.

### DIFF
--- a/app-starter/WguAppTemplate.vue
+++ b/app-starter/WguAppTemplate.vue
@@ -40,8 +40,8 @@
         <!-- layer loading indicator -->
         <wgu-maploading-status />
         <slot name="wgu-after-map" />
-        <!-- Portal to overlay the map content from an application module -->
-        <portal-target name="map-overlay" />
+        <!-- Teleport to overlay the map content from an application module -->
+        <div id="wgu-map-teleport" />
         <wgu-app-logo />
         <wgu-bglayerswitcher />
         <wgu-overviewmap v-if="overviewMapConfig" v-bind="overviewMapConfig"/>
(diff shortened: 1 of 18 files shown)


fschmenger:
Hi guys, it´s been a while but I finally had the time to test the recent Wegue version with a larger custom application. Most parts worked out fine instantly - a great job!  - This PR fixes some mostly simple oversights.

Regarding 128395bb611f4751e43fd5c3661ff18012271e93: This introduces some on-* colors that we could use in Vue2 based versions and isn`t exactly necessary from a bugfix point of view. However it was required to provide the fix in b58cccd424757a64117f340d83788fec308c29ea. I hope you don't mind that I put those in here.

I have another branch setup with 2-3 simple layout related tweaks. These aren`t functional so I will post them seperately.
Greets, Felix

sronveaux (approved):
Hi @fschmenger and thanks for your return on the latest changes and for this PR !

For sure, this was a lot of work and it is difficult to think about all use cases and verify all this. So this kind of returns are more than welcome, even more when coming from people who already know the product and used it quite a lot !

I think all this makes sense and is really interesting. I've just made some simple suggestions with missing semicolons which are purely for esthetics, nothing important there.

Just to give some information about some of your changes and keep a log somewhere:
- Vue-portal V3 was still used while migrating to Vue3 as the Teleport component was still flagged as experimental at that time.
- Good catch with Vuetify locales import. That's one thing that was migrated and worked at a certain time, but as Vuetify was updated a few times during the whole process, I'm pretty sure some other glitches may have appeared unfortunately.
- No problem at all to add back some `on` styles with Vuetify colors, that's a point on which upgrading was really a problem. Those styles still exists and are automatically assigned in some circumstances (as you certainly saw) but you can't reference them easily with auto-defined classes. Adding them back can be handy in some circumstances then.

For the rest, nice catches and additions, thanks again for the good work !

I already approved the PR, so merge at will, except if @chrismayer has some comments to add somewhere...

Cheers, Sébastien

PR #455, opened 2025-11-11
Restrict the measure tool draw interaction to the left mouse button. 

### LINKED ISSUE #438: Measure Tool: Computation not taking place when drawing with middle and right mouse button
Once the `Measure Tool` is open, you can draw lines and/or polygons for their length, area and angle to be computed.

You can use any mouse button to draw, but only shapes drawn with the left button will trigger the computation.

It would either have to compute sizes regardless of which mouse button is used, or only draw when the left button is used, in order to remain consistent.

fschmenger:
This pull request restricts the draw interaction of the measure tool to allow selections with the left mouse button only - closes #438.

### DIFF
--- a/src/components/measuretool/OlMeasureController.js
+++ b/src/components/measuretool/OlMeasureController.js
@@ -74,8 +74,11 @@ export default class OlMeasureController {
     }
 
     const type = (measureType === 'area' ? 'Polygon' : 'LineString');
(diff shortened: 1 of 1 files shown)


chrismayer (approved):
Thanks for this @fschmenger. Tested it and works well. Also on mobile phone.

Feel free to merge...

PR #458, opened 2025-11-17
Updated locale files for ArcGIS test layer.

fschmenger:
This PR updates the translations for ArcGIS test layer to avoid console warnings due to missing language keys.


sronveaux:
And can confirm for the French translation that `Padova` can be called `Padoue`.

Issue #457, opened 2025-11-12
Make vite application base path relative?


fschmenger:
In one of my Wegue based apps I'm using a reverse proxy to serve the application from `https://mydomain/subpath` which maps to the root path of the NGINX based docker image - with Wegue default settings. The proxy rules are fairly standard.
```
ProxyPass        /subpath http://nginx_client
ProxyPassReverse /subpath http://nginx_client
```
Now per default vite is configured to emit absolute paths like 
```
<script type="module" crossorigin src="/assets/main-BTj3Mhow.js"></script>
<link rel="stylesheet" crossorigin href="/assets/main-Gd_3wRf1.css">
```
and subsequently the path resolution fails, as the webserver is trying to resolve `https://mydomain/assets/main-BTj3Mhow.js` (`instead of https://mydomain/subpath/assets/main-BTj3Mhow.js`)

A previous vue-CLI based built resulted in emitting relative paths, which worked out of the box, e.g.
 ```
<script type="module" crossorigin src="./assets/main-CA9XZd7C.js"></script>
<link rel="stylesheet" crossorigin href="./assets/main-Dw7UkqbL.css">

```

I'm aware that I could set the env variable `WGU_PUBLIC_PATH` to get the desired behavior (or I could introduce some ugly rewrites on the reverse proxy side).
What I´m asking myself:

Would it be feasible and better to configure the path in `vite.config.js` like
` base: (process.env.NODE_ENV !== 'test' && env?.WGU_PUBLIC_PATH) || './'`
to be relative, so we can retain backward compatibility?
And similarly should we set `ARG PUBLIC_PATH=./` in the `Dockerfile`?
I tested this setting and it worked out of the box while the user can still customize it and override `WGU_PUBLIC_PATH`.

PR #460, opened 2025-11-25
Make Vite application default base path relative

sronveaux:
Very simple PR that follows discussion #457.

Default base path is set to a relative path and a note is added to recommend users to define the `WGU_PUBLIC_PATH` variable if deployment configuration is known in advance.


fschmenger:
Nice that you arrived at the same commit that I used in one of my apps - so I don`t have to test it :)
Thanks for the fix!

fschmenger (approved):

PR #464, opened 2025-12-08
Pin font icons versions to latest

sronveaux:
This small PR pins the `Material Design Icons` and `material-icons` to latest available versions.  
This is part of the must-have points to address before `Wegue v3` is out and is thus mentioned in the upgrade notes.

The PR does nothing in practice, as the versions present inside the package.json file were:

- "@mdi/font": "^7.2.96"
- "material-icons": "^1.13.9"

As latest versions are `7.4.47` for `MDI` and `1.13.14` for `material-icons` and as dependencies were pinned with a `^`, someone who had installed dependencies lately was already fetching those latest versions.  
However, as discussed in #319 some times ago, changing version of those icon fonts can bring bad surprises as some icons are removed or replaced from version to version. In our case, no icon disappeared but some were updated and so can have a different visual even between minor version updates. That's why I propose to pin those fonts to the exact version as is already the case for `OpenLayers` and `Proj4`.

I also propose to move them from the `devDependencies` to the `dependencies` section as they are clearly part of the assets distributed at runtime and are not dev-only. In practice, as we're using `Vite`, the change has no impact, it is purely philosophical and merely clarifies the use that is made of these packages.


chrismayer:
Totally agree that these are **no** `devDependecies`.

Regarding pinpointing: If we can assure that semantic versioning is used by these 2 packages we could use `~` or `^`,  otherwise I am also fine with pinning them.

PR #466, opened 2025-12-09
Replace no more existing "gas-wfs" in example application 

chrismayer:
This replaces the no more existing `"gas-wfs"` based on the WFS at ows.terrestris.de with another open WFS by German BKG ("Federal Agency for Cartography and Geodesy") delivering airports in Germany:

image

This only affects some example applications.

### DIFF
--- a/docs/wegue-configuration.md
+++ b/docs/wegue-configuration.md
@@ -521,13 +521,14 @@ Example configurations can be found in the `app-starter/static` directory. Below
       "loadOnlyVisible": true,
       "visible": false,
       "selectable": true,
+      "columnMapping": {
+        "name": "Name",
+        "poi_id": "POI ID"
+      },
       "style": {
         "textIcon": "airplane_ticket",
         "font": "normal 30px Material Icons",
         "fillColor": "black"
-      },
-      "columnMapping": {
-        "name": "Name"
       },
       "selectStyle": {
         "textIcon": "airplane_ticket",
--- a/app-starter/public/static/app-conf.json
+++ b/app-starter/public/static/app-conf.json
@@ -103,6 +103,10 @@
       "loadOnlyVisible": true,
       "visible": false,
       "selectable": true,
+      "columnMapping": {
+        "name": "Name",
+        "poi_id": "POI ID"
+      },
       "style": {
         "textIcon": "airplane_ticket",
         "font": "normal 30px Material Icons",


docs/wegue-configuration.md
```diff
@@ -511,29 +511,28 @@ Example configurations can be found in the `app-starter/static` directory. Below
...
       "style": {
-        "textIcon": "local_gas_station",
+        "textIcon": "airplane_ticket",
         "font": "normal 30px Material Icons",
-        "fillColor": "blue"
+        "fillColor": "black"
       },
       "columnMapping": {
         "name": "Name"
       },
```
sronveaux: Depending on the answer of the other comment up here, should this be amended or not ?  Completely open question...

sronveaux (commented):
Hi @chrismayer,

Thanks for tackling this and providing a seemingly stable source for the future.

For sure `gas-wfs` was a nice example as it had points all over the world and its interactions with the `AttributeTable` and its dynamic loading could be directly evaluated.  
I also liked the one @fschmenger proposed with cities all around the world. However, pinning airports is visually easier and more talkative than pinning cities I think.

Simple question concerning `docs/maps-layer-configuration.md`, line 156 which looks like this:

>  | textIcon           | Point only, see [text](https://openlayers.org/en/latest/apidoc/module-ol_style_Text-Text.html). Icons for font `normal 30px Material Icons` can be found [here](https://fonts.google.com/icons?selected=Material+Icons) | `"textIcon": "local_gas_station"` |

Should we still reference the now unused `local_gas_station` or not ?  It can certainly stay as is, this is a valid example. I just wanted to be sure it wasn't "missed"...

I also added another comment on the config file to be discussed fast before merging but I really like it, thanks!

chrismayer:
Thanks for your review @sronveaux :+1: 

I re-added the columnMapping since it was quite messy in the AttributeTable without it. Regarding 

> I also liked the one @fschmenger proposed with cities all around the world

I tried this WFS but stumbled over some problems with the service, which I could not fix/correct. But changing to another WFS in a later stage shouldn't be a big deal.

> Simple question concerning docs/maps-layer-configuration.md, line 156 which looks like this

I decided to keep the "old" icon, since it is valid and gives some diversity.

I am going to merge once the test suite is green, which fails for `GeoCoderController calls remote service`, which is surely totally unrelated.

PR #477, opened 2026-04-02
Extend jsconfig.json for better Visual Studio Code experience.

fschmenger:
This extends `jsconfig.json` to improve IntelliSense and navigation behavior of VisualStudio (and potentially other IDEs using a JavaScript language service). For a documentation of available options see https://code.visualstudio.com/docs/languages/jsconfig.

Some brief explanation:
* Enable type-check for JavaScript files: `"checkJs": true`: 
* Improve reliability of path resolution
```
"baseUrl": ".",
"paths": {
  "@/*": ["src/*"],
  "APP/*": ["app/*"]
} 
```
* Add explicit includes: `"include": [....]`
* Tell VSCode how to handle module resolution: `"moduleResolution": "Node"`
* Help VSCode to assume ESM syntax and understand JS features: 
```
"module": "ESNext"
"target": "ES2020"
```

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,9 +1,26 @@
 {
   "compilerOptions": {
+    "target": "ES2020",
+    "module": "ESNext",
+    "moduleResolution": "Node",
+    "checkJs": true,
+
+    "baseUrl": ".",
+
     "paths": {
-      "@/*": ["./src/*"],
-      "APP/*": ["./app/*"]
+      "@/*": ["src/*"],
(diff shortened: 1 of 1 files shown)


fschmenger:
This gives alot better suggestions for autocompletion, when the type can be deduced. i,E for local variables and expressions.
Here and there the syntax check is complaining. We should address that in the future, e.g.
* `$appConfig` on the Vue prototype is unknown. 
* A revision of the `Map` component, to properly declare all data properties

PR #478, opened 2026-04-08
Upgrade vite dependency to 7.3.2

chrismayer:
Upgrades `vite` to version `7.3.2`, which adds a vulnerability patch, see [here](https://github.com/vitejs/vite/security/advisories/GHSA-p9ff-h696-f583).


fschmenger (approved):
Thanks Christian!
I tested hot module reload, build, preview and unit tests. Everything works as expected.

PR #481, opened 2026-04-23
Layer legend wms styles support

fschmenger:
The following PR extends the layer legend implementation to forward the WMS `STYLES` parameter into `GetLegendGraphic `requests. Reactivity was added, such that the legend image is re-rendered when new styles are assigned to the WMS source.


chrismayer (approved):
Thanks for this @fschmenger. Very useful extension. LGTM, also tested in example apps.

fschmenger:
Hi @chrismayer. Thanks for the review - will merge now.

PR #484, opened 2026-05-11
Do not reverse the list of map layers.

fschmenger:
This PR removes reversing of layer collections during both OpenLayers map initialization and layer list rendering. Layer order is now preserved consistently as declared in configuration files and runtime additions. This has the following practical consequences:
* Rendering z-order now follows native OpenLayers behavior (later layers render above earlier ones when sharing the same z-index)
* Dynamically added layers during runtime (e.g. DragDrop) now appear at the end of layer selection and management controls instead of at the top.

### DIFF
--- a/src/components/attributeTable/AttributeTableWin.vue
+++ b/src/components/attributeTable/AttributeTableWin.vue
@@ -72,8 +72,7 @@ export default {
           layer instanceof VectorLayer &&
           layer.get('lid') !== 'wgu-measure-layer' &&
           layer.get('lid') !== 'wgu-geolocator-layer'
-        )
-        .reverse();
+        );
     },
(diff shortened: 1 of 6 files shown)


fschmenger:
The logic here was a bit twisted and a bit counterintuitive. I'm not sure why it was implemented this way. I assume this dates back to the days when we neither had a distinction between foreground and background layers, nor support for z-index in configurations. 
Currently this behavior hits me in one of my apps, as I`m creating a bunch of dynamic layers, and I want the ones being created last to show up at the end of LayerList.

However, you can argue that the UI should show topmost layers on top of lists (This would however required proper sorting by z-index).

Any thoughts and insights welcome.

